PR #9, opened 2017-05-04
Added cli option to allow no prefix for tag attribute

eric2036:
For my project i needed tag attribute without any default prefix like "@_" so i have forked this project and modified cli.js to allow no prefix for tag attribute through  cli option as "-np" (no prefix) .  

Usage before :
$xml2js [-ns|-a|] <filename> [-o outputfile.json]

Usage After :
$xml2js [-ns|-a|-np] <filename> [-o outputfile.json]

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -27,9 +27,11 @@ if(process.argv[2] === "--help" || process.argv[2] === "-h"){
             options.ignoreTextNodeAttr = true;
         }else if(process.argv[i] === "-o"){
             outputFileName = process.argv[++i];
+        }else if(process.argv[i] === "-np"){ // no prefix for attribute
+            options.attrPrefix = "";
         }else{//filename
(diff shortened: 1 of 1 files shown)


amitguptagwl:
@eric2036 Thanks for the changes you suggested and your effort for the implementation. But I'll not be able to merge the changes and publish them because of following reasons;

1. I dint write the unit tests for CLI feature yet. So I can't check for any bug quickly. However I can't see any bug but there are small changes which are required to be taken care. Eg. console log on "--help" option should be updated.
2. In my opinion, a CLI argument should be added not to remove the default prefix but to override it. Or default prefix for attribute name, in case of CLI, should be set to blank and some CLI argument should be introduced to set prefix value. (This is a good point to discuss/review)

Unfortunately, I have no machine for next 2 weeks to do any change.

Issue #36, opened 2018-01-24
Since 2.9.*, that XML became invalid


j0k3r:
Since the 2.9.0 release, the below xml become invalid when using `xmlParser.validate(xml)`.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE daily_western_horo_fr>
<daily_western_horo_fr>
   <date valeur="2018-01-24">
      <signe valeur="vierge">
         <amour>La vie de couple s'accompagne de tâches pas toujours très agréables à réaliser. Aujourd'hui, cela ne vous gênera pas. Qu'il s'agisse de ménage, de cuisine ou de gestion du foyer, vous trouverez cela normal et serez heureux de le faire, car vous vous sentez responsable de votre petit monde. Vous vous montrerez adorable quoi qu'il arrive.</amour>
         <argent>Ah ! Les imprévus ! Que vous n'aimez pas ça ! Et pourtant, c'est ce qui risque d'arriver aujourd'hui, en raison d'influx dissonants Lune/Uranus dans le secteur concerné. Rassurez-vous, votre gestion impeccable devrait vous permettre d'assurer sans problème.</argent>
         <sante>Vous canaliserez à merveille votre énergie. Votre esprit est pratique et vous possédez un excellent sens de l'organisation. Tout roule !</sante>
         <travail>Toujours soucieux de rendre un travail impeccable, vous serez en mode perfection. Comme la perfection n'existe pas, vous essaierez au moins de vous en rapprocher le plus possible. Vous perdrez du temps pour des détails que personne ne remarquera...</travail>
         <famille_foyer>Vous serez heureux de partager de petits plaisirs simples et spontanés avec les enfants comme avec les plus grands. L'ambiance se fera très complice.</famille_foyer>
         <vie_sociale>Vos amis seront tentés de faire appel à vos capacités de jugement. Ils savent que vous ne laissez rien au hasard et que lorsque vous affirmez quelque chose, ils peuvent vous faire confiance.</vie_sociale>
         <citation>« L'absence est le plus grand des maux. » La Fontaine</citation>
         <nombre_chance>23</nombre_chance>
         <clin_doeil>Vous serez très ouvert à de nouvelles sensations, mais en aurez-vous l'opportunité ? Si vous n'êtes pas très clair sur vos intentions, vous passerez à côté d'une occasion.</clin_doeil>
      </signe>
   </date>
</daily_western_horo_fr>
```
[...]

PR #37, opened 2018-01-24
Correctly remove CRLF line breaks

j0k3r:
Related to https://github.com/NaturalIntelligence/fast-xml-parser/issues/36

In fact, the XML I retrieve from an external API has CRLF line break instead of just LF.
Your regex doesn't handle that case that's why the validator always return `false` (see https://stackoverflow.com/a/10805198/569101)

### DIFF
--- a/bin/validator.js
+++ b/bin/validator.js
@@ -3,7 +3,7 @@ var util = require("./util");
 
 var tagsPattern = new RegExp("<\\/?([\\w:\\-_\.]+)\\s*\/?>","g");
 exports.validate = function(xmlData){
-    xmlData = xmlData.replace(/\n/g,"");//make it single line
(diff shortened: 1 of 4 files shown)


amitguptagwl:
Thanks for the changes and tests. I completely agree that I didn't consider '\r' in my regex.

Delagen:
@amitguptagwl Any plan to update NPM?

amitguptagwl:
:) I was actually busy with some other project. But I'll publish it tomorrow morning sure.

Delagen:
@amitguptagwl Thanks. Temporary use git version )) Also reach this issue when parse SOAP responses from windows service )

amitguptagwl:
Done. published :)

Issue #40, opened 2018-02-02
Decode HTML entities in text values


Delagen:
### Checklist

* [x] Are you running the latest version?
* [x] Include sample input XML 
* [x] Include actual output
* [x] Include expected output
* [x] Did you try online [tool](https://naturalintelligence.github.io/fast-xml-parser/)?
* [x] Did you [star](https://github.com/NaturalIntelligence/fast-xml-parser/stargazers) the repository for further updates? ;)

I include PR soon

amitguptagwl:
Published

amitguptagwl:
I believe it should not parse HTML entities withing CDATA range.

Delagen:
@amitguptagwl CDATA also can contain HTML entities

amitguptagwl:
Yes. :) 
So what I expect from `<tag><![CDATA[&lt;sender&gt;John Smith&lt;/sender&gt;]]></tag>` is 

```
{
   "tag" : "&lt;sender&gt;John Smith&lt;/sender&gt;"
}
```

But what I'm getting from current parser is

```
{
   "tag" : "<sender>John Smith</sender>"
}
```

Delagen:
> But what I'm getting from current parser is

And it's right

amitguptagwl:
As per the [wiki](https://en.wikipedia.org/wiki/CDATA),

> Character data is character data, regardless of whether it is expressed via a CDATA section or ordinary markup. CDATA sections are useful for writing XML code as text data within an XML document.

 > if the numeric character reference `&#240;` appears in element content, it will be interpreted as the single Unicode character 00F0 (small letter eth). But if the same appears in a CDATA section, it will be parsed as six characters: ampersand, hash mark, digit 2, digit 4, digit 0, semicolon.

Above phrase in my words (understanding), 

> if `&#240;` (represents to &#240;) appears in CDATA section should not be parsed to &#240; but to `&#240;`  only.

Delagen:
@amitguptagwl I will make some refactoring, and place PR tomorrow

amitguptagwl:
Thanks @Delagen . In case if you are busy with some other prior work, we can just roll back the changes for the time being. And will handle them later. I'm anyhow planning to rewrite parser but it'll take at least a month :smile_cat:

Delagen:
Thanks @amitguptagwl for work, I placed PR to not decode CDATA
But seems parser ignore sibling text when CDATA present
```
<a>
[...]

PR #42, opened 2018-02-02
fix(entities): decode only text in attribute values and non CDATA text, updates #40

Delagen:


amitguptagwl:
Thanks for the changes you made. But I'm not able to merge the changes and closing this PR because the parser have been modified to handle the CDATA and tag data separately.

There are few users who don't want decoding. So I have made it optional in v3.

Delagen:
@amitguptagwl That's OK. Thanks.

PR #57, opened 2018-02-27
fix namespaces can contain the same characters as xml names

Delagen:


amitguptagwl:
long time to see you :wink:

Delagen:
My co-worker pushes me sometimes for testing this in production )

PR #63, opened 2018-03-02
Keep undefined nodes out of the XML output.

HaroldPutman:
This is useful when you are deleting nodes from the JSON and rewriting XML. While I could see a case for writing an empty node in this case (`<b/>`), I'm fairly certain we never want the current behavior which is `<b>undefined</b>`.

so if you have:
```
const obj = {
  house: {
    pet: [
        "cat",
       "spider",
       "dog"]
   }
};
delete obj.house.pet[1]; // get rid of spider!
var result = parser.parse(obj);
```
You get:
```
<house>
  <pet>cat</pet>
  <pet>undefined</pet>
  <pet>dog</pet>
</house>
```
IMO you wanted:
```
<house>
  <pet>cat</pet>
  <pet>dog</pet>
</house>
```

### DIFF
--- a/spec/j2x_spec.js
+++ b/spec/j2x_spec.js
@@ -50,6 +50,24 @@ describe("XMLParser", function () {
         expect(result).toEqual(expected);
       });
 
+      it("should supress undefined nodes", function () {
+        var jObj = {
+            a : {
+                b : [
+                    undefined,
+                    "val1",
+                    { c : "val2",
+                      d : undefined,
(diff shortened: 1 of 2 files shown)


HaroldPutman:
(Sorry my editor also got rid of trailing whitespace on some lines)

amitguptagwl:
Sorry for the late response, it's a festive season here. I will check and merge it on Monday.

PR #95, opened 2018-06-15
Update README.md

lfplazas10:
For someone who hasn't used the 'he' library before, it's quite confusing looking at the readme and trying to implement the examples if you don't know what 'he' is, it kinda looks more like an undeclared variable. It would be helpful to make this clarification for potential users.


amitguptagwl:
I completely agree with you.

PR #100, opened 2018-07-16
added missing TypeScript definition for j2xParser

driehle:
# Purpose / Goal
Currently the usage of fast-xml-parser with TypeScript is only limited when converting JSON into XML. This is because the TypeScript definition of j2xParser is missing. This PR adds the missing definition.

# Type
[X]Bug Fix
[   ]Refactoring / Technology upgrade
[   ]New Feature


amitguptagwl:
Thanks for the PR

Issue #104, opened 2018-08-15
invalid whitespace inside xml tag when format option is true and nodes have attr text


juxe:
### Checklist

* [X] Are you running the latest version?
* [X] Have you included sample input, output, error, and expected output? 
* [X] Have you checked if you are using correct configuration?
* [X] Did you try online [tool](https://naturalintelligence.github.io/fast-xml-parser/)?

### Input
```json
{
  "container": {
    "body": {
      "my-node": {
        "#text": "foo",
        "@_staticMessage": "bar"
      }
    }
  }
}
```

### Code
var json is the above input
```js
function js2xml (json) {
  const Parser = fastXmlParser.j2xParser

  const parser = new Parser({
    ignoreAttributes: false,
    format: true
  })
  return parser.parse(json)
}
```

### Output
```xml
<container>
  <body>
    <my-node staticMessage="bar">
foo    </my-node>
  </body>
</container>
```
### expected data
```xml
<container>
  <body>
    <my-node staticMessage="bar">foo</my-node>
  </body>
</container>
```

**Would you like to work on this issue?**

- [X] Yes
- [ ] No

[Bookmark](https://github.com/NaturalIntelligence/fast-xml-parser/stargazers) this repository for further updates.

PR #105, opened 2018-08-15
Fix formatting for nodes with attributes

juxe:
# Purpose / Goal
I would like to fix a bug that I encountered when parsing json to xml with the format option flag set to true. If the JSON doc specifies nodes with an attr string and text value, an extra newline is added between the > of the open tag and the node text value, and extra whitespace is added to the end of the text value equal to n * level (where n is the number of spaces in indentBy, default 2).

https://github.com/NaturalIntelligence/fast-xml-parser/issues/104

# Type
Please mention the type of PR

[X]Bug Fix
[ ]New Feature

### DIFF
--- a/src/json2xml.js
+++ b/src/json2xml.js
@@ -168,7 +168,7 @@ function replaceCDATAarr(str, cdata) {
 function buildObjectNode(val, key, attrStr, level) {
     return this.indentate(level)
            + "<" + key + attrStr
-           + this.tagEndChar
+           + ">"
            + val
            //+ this.newLine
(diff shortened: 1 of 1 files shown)


amitguptagwl:
Thanks for your PR. will you mind to add a test for that?

juxe:
ok, so I updated one of the current tests that was very similar to what I was writing, and it is working and passing tests now, but I'm hitting a block getting codeclimate to accept it. I'm sure there's a way to refactor it a bit for codeclimate to pass, but I'm not seeing it at the moment

PR #109, opened 2018-08-21
feat(bundle): migrate to webpack

Delagen:
# Purpose / Goal
Simple make split bundles with entries in webpack.config.js
Use latest features of ECMAScript language without any hacks

# Type
Please mention the type of PR

[ ]Bug Fix
[x]Refactoring / Technology upgrade
[ ]New Feature


Delagen:
I think to remove outdated node versions. They all reach EOL )

amitguptagwl:
Thanks @Delagen . Though moving to webpack will enable us splitting this package in small browser bundles, using ECMAScript latest syntax in the code  etc, I wanted to check few things before merging this PR

1. Need of generating *parser.node.js*.
2. Need of pointing main script to *lib/parser.node.js*.
3. Impact on typescript definitions.
4. impact on vscode debugging

Delagen:
1. Cause NodeJS can miss some function babel can provide
2. Because of 1
3. No impact
4. Don't know

amitguptagwl:
> Cause NodeJS can miss some function babel can provide

For example? are we using any such function? As this module is pretty stable I see less features to be implemented in the future.

Delagen:
For example ES2015 modules, async/await by default or something else... Need to check tables

amitguptagwl:
async/await is supported by node js. And we're not using that in our module anywhere. Can you please share some reference link where I can check those features? I couldn't find it through google.

Delagen:
https://node.green/

Delagen:
Many browser modules use ES2015 module model for efficient tree shaking

Delagen:
I don't think that NodeJS bundle really needed at this time, but creating it can offer support more older versions, but with lack of some travis testing, or we can make some tweaks )

Delagen:
I you argue for removing it, I can fix it )

amitguptagwl:
@Delagen With my own experience and under the guidance of my supervisors I've learnt that **Plan for future but code for current**

Delagen:
Current is different for someone ) Some enterprise projects can use 0.12.x version or even older )
So should I remove this node bundle?

amitguptagwl:
Yes please

.travis.yml
```diff
@@ -1,11 +1,8 @@
 language: node_js
 node_js:
-  - 4
-  - 5
   - 6
-  - 7
   - 8
-  - 9
+  - 10
```
amitguptagwl: Any particular reason to remove 7, 9?

Delagen:
Then I think we need to create some entry points for building multiple bundles

amitguptagwl:
Yes. And I'm thinking to remove XML to Nimn transformation as Nimn seems duplicate of Google's protobuf.

Zalastax:
This might be too late since it's already merged, but shouldn't rollup have been preferred over webpack, given that this is a library? Webpack has some overhead that rollup doesn't is my concern. It shouldn't be difficult to switch in the future though so might be best to keep with the flow you have going.

amitguptagwl:
@Zalastax Thanks . It seems interesting. We can have a look.

Issue #35, opened 2018-01-24
Creating tag/release on GitHub each time you release a new version on NPM


j0k3r:
My project got some validation problem on some XML using 2.9.* version of your library.
2.8.* is working great.

Since you didn't create tag or release on GitHub I can't compare what changed on your side to understand why parsing is failing.

Could you create tag / release on GitHub each time you release a new version on NPM? At least for future release because I guess it'll be hard for previous release.

Thanks.

amitguptagwl:
Thanks, it's a good suggestion. Even I thought about it before 2nd last release but missed because I got stuck with other work. 

I really feel sorry if I made your build red. I'll definitely create the changelogs soon. For the time being you can check the commit history, or you can tell me what is basically breaking. So I may help to figure out the issue.

j0k3r:
I'll open a different issue about the problem :wink:

amitguptagwl:
Yeah sure :)

amitguptagwl:
Changelog added

j0k3r:
You still haven't create a release or a tag on GitHub: https://github.com/NaturalIntelligence/fast-xml-parser/releases

amitguptagwl:
I read it somewhere that git automatically tag npm repo as per version mentioned in package.json. Let me check that if it doesn't work

amitguptagwl:
Check if this work for you git@github.com:NaturalIntelligence/fast-xml-parser.git#v2.9.0

j0k3r:
Well you should create a tag on git and then publish it to GitHub and it'll appear here: https://github.com/NaturalIntelligence/fast-xml-parser/tags

You can choose to:
- manually create a release on GitHub each time you publish to npm: https://help.github.com/articles/creating-releases/
- manually create a tag on your git repo and then publish it to GitHub: https://git-scm.com/book/en/v2/Git-Basics-Tagging

amitguptagwl:
Thanks @j0k3r .  Yes it can be good practice for github users/visitors. But will do it next time.

j0k3r:
Great, you can still create a release right now and name it 2.9.3 :wink: 
Also, you forgot the 2.9.3 information in the changelog

amitguptagwl:
You'll not let me lazy :laughing:

robations:
@amitguptagwl If you run `npm version x.y.z` it creates a commit that updates package.json with the new version number, and also tags that commit.

[...]

PR #118, opened 2018-09-28
Few update to README

j0k3r:
- fix bad importer parser (use of `fastXmlParser` instead of `parser`)
- fix indentation
- remove the way to use `parser.js` directly from GitHub (it is served as `text/plain` and can't be parsed by browsers)

Also, please, **please**, :pray:, publish a changelog with **breaking changes** from one version to an other. Or maybe create an UPGRADE.md file to tell people how to upgrade from a previous major version.

It's still related to that issue https://github.com/NaturalIntelligence/fast-xml-parser/issues/35

I spend few hours fixing code while uprgading from 2.8.3.

Some options got renamed (`attrPrefix` => `attributeNamePrefix`), some options disapear (like `ignoreNonTextNodeAttr`), etc.
Also `validate()` function does not return `false` but an object now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ To use it from **CLI** Install it globally with `-g` option.
 
 `$npm install fast-xml-parser -g`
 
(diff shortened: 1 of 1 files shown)


amitguptagwl:
As I remember, I highlighted the broken changes on README.md and removed them after few months. But yes we can maintain UPGRADE.md to highlight such changes.

Applogies for your trouble.

PR #119, opened 2018-10-11
Add inline empty tag functionality when format is set to true

flaviocolonnascytl:
# Purpose / Goal
The purpose of this edit is to allow to the developer to choose how an empty tag should be rendered.
In this case, setting a configuration variable to true, it is possible to render it as non self-closed tag.
Here below you'll find an example:
- Current situation: `<myEmptyTag />`;
- Setting the variable: `<myEmptyTag></myEmptyTag>`

This is useful when you have to check an xml signature and you want to create the xml from the json structure, according to your format.

# Type
It is a new feature.

[ ]Bug Fix
[x]New Feature


amitguptagwl:
Can you please mention the purpose and add necessary tests for this PR?

flaviocolonnascytl:
I will add more details once finished.

amitguptagwl:
@flaviocolonnascytl Any update on this PR?

flaviocolonnascytl:
I have updated the goal and why we need it.
Further, I will include new tests.

amitguptagwl:
@flaviocolonnascytl  I hope you must have read the README completely before this change. And you must have noticed `supressEmptyNode` option. It basically helps a user to decide how an empty tag should be rendered.

Please update the purpose/goal, if you have different aim.

flaviocolonnascytl:
Hi,

I read it and my purpose is different. I was meaning HOW it should be rendered and not if SHOULD be rendered.
You supressEmptyNode just delete it but I want to change how it is rendered.

amitguptagwl:
Have you tested with `supressEmptyNode` option?

flaviocolonnascytl:
Yes of course. But the suppressEmptyNode option just render an empty tag like this:
<myEmptyTag />.
What we want is to render it as following:
<myEmptyTag></myEmptyTag>

amitguptagwl:
Either I'm failing to understand your point or to explain mine. Currently

* when `suppressEmptyNode` is set to `true`
```xml
<root>
    <element>1</element>
    <element2>
        <subelement aaa="sub_aaa" />
    </ element2>
    <date>test</date>
</root>
```

* when `suppressEmptyNode` is set to `false`
```xml
<root>
    <element>1</element>
    <element2>
        <subelement aaa="sub_aaa"></subelement>
    </element2>
    <date>test</date>
</root>
```

Notice the `subelement` tag in the above example. You can't delete the tag but can decide if it should be rendered as a self-closing tag or a tag with the closing tag.

I'm not able to understand how is your point different from this.

amitguptagwl:
@flaviocolonnascytl I would be happy to clear the doubts.

amitguptagwl:
Considering that this is the duplicate feature, Can we close this PR? I would like to hear back your views.

flaviocolonnascytl:
The problem was when indentation was on. If you indentate an empty tag, it automatically is rendered as following:
```
<myEmptyTag>
</myEmptyTag>
```
Meanwhile, I want it to be rendered inline.

amitguptagwl:
Though I haven't tested that. It looks like you need a bug fix. And I'll suggest you raise a fresh PR for that.

Issue #79, opened 2018-04-26
Add feature of JSON to XML to online tool 


amitguptagwl:
> This issue is raised for developers who are just starting with open source development. So please refrain yourself if you're experienced.
 
Please check README or [spec](https://github.com/NaturalIntelligence/fast-xml-parser/tree/master/spec) folder for examples.

[Online demo tool](https://naturalintelligence.github.io/fast-xml-parser/) demonstrate what XML parser can do. However it is currently just parsing XML to JSON and validating XML only. As a part of this change, we have to parse JSON to XML;

You may discuss here about design changes.

Check [how to](https://github.com/Roshanjossey/first-contributions).

geekiam23:
I'd like to give it a shot.

amitguptagwl:
@geekiam23 any update on this issue? Is it fine if any other contributor can pick this issue?

geekiam23:
@amitguptagwl Yeah, I couldn't figure it out. Please let someone else give it a shot.

amitguptagwl:
This issue can be picked by any contributor now. Demo UI has been simplified. So now it is just the matter of 10 mins change.

sakshivij:
Hey! I would like to work on this issue.

amitguptagwl:
Yes, please go on.

amitguptagwl:
@sakshivij have you got the chance to work on?

sakshivij:
Yes! I am working on it. I just need this weekend. If I'm not able to resolve the issue, I'll let you know by Sunday.

amitguptagwl:
No problem, take your time. It's not urgent. I just wanted to check if you have stuck in some issues and need help

sakshivij:
A little bit of hint would be great!

amitguptagwl:
Umm.. what are you actually trying?

1. Open the index.html
2. Open the section which is showing options and buttons to parse XML to JSON
3. Paste and create another section. Change the IDs. Modify the options as per the example snippet given on [README.md](https://github.com/NaturalIntelligence/fast-xml-parser#json--js-object-to-xml)
4. Add javascript code to read options and call JSON to XML parser. It is present on index.html only.

sakshivij:
Yes! I have done the same thing. But the code gives error. 
![image](https://user-images.githubusercontent.com/20201675/42135251-0769a274-7d66-11e8-92b3-ada697a65e99.png)

sakshivij:
[...]

PR #127, opened 2018-11-08
Changes on index.html

lfplazas10:
While I was working on #79 I noticed that if the user changes something on the result textarea, any further Convert or Validate calls did not update the textarea content. The solution for this was quite simple, just had to change all $.text() calls to $.val()

### DIFF
--- a/index.html
+++ b/index.html
@@ -70,33 +70,33 @@
             });
 
             $('#toJson').on('click', function() {
-                $('#result').empty();
+                $('#result').val('');
                 var result = parser.parse($('#txtXml').val(),buildParsingConfig());
                 updateOutputLength(JSON.stringify(result));
-                $('#result').text(JSON.stringify(result,null,4));
+                $('#result').val(JSON.stringify(result,null,4));
             });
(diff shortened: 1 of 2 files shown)


lfplazas10:
Updated the demoapp_spec.js file as it didn't recognize the formatted content. I see all tests are passing but there seems to be something not working with the security validator.

lfplazas10:
The CI tool runs 3 tasks, I can see there is one called security-something that fails, but seems to be a problem with the tool itself.

amitguptagwl:
@lfplazas10 Sorry but I'm not able to link your words. I couldn't find anything like "security*" in the whole repository. Can you please share the screenshot? 

I believe you're separately working on JSOn to XML parsing demo.

lfplazas10:
screen shot 2018-11-08 at 11 11 00 pm

I was talking about the security/snyk check performed by the CI tool.

amitguptagwl:
Ah ok. I got your point. Thanks. That's not an issue. So ignore that.

Issue #129, opened 2018-11-23
XML fails validation if it contains an attribute called "length"


brainsiq:
If you try to validate an XML document that has a "length" attribute anywhere it fails with an error "attribute length is repeated". I've tracked this down to the use of an array [here in the validator](https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/src/validator.js#L253). The code that checks for repeated attributes [checks if length is a property of the array](https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/src/validator.js#L270). It appears that `attrNames` can just be initialised as an object instead of an array to resolve this issue.

### Checklist

* [x] Are you running the latest version?
* [x] Have you included sample input, output, error, and expected output? 
* [x] Have you checked if you are using correct configuration?
* [x] Did you try online [tool](https://naturalintelligence.github.io/fast-xml-parser/)?

 
### Input
`<test length="1"></test>`

### Code
```js
require('fast-xml-parser').validate('<test length="1"></test>');
```

### Output
`{ err: { code: 'InvalidAttr', msg: 'attribute length is repeated.' } }`

### expected data
`true`

**Would you like to work on this issue?**

- [x] Yes
- [ ] No

PR #130, opened 2018-11-23
Pass validation if XML contains 'length' attribute

brainsiq:
# Purpose / Goal
Bugfix for #129 

When validating XML containing a 'length' attribute currently an error is always returned 'attribute length is repeated.'. 

This is because the code uses an array to check for repeated attributes and calls hasOwnProperty('length'). I have changed this to use an object instead of an array so hasOwnProperty('length') will only return true when there are multiple length attributes. The test I have added verifies this.

# Type

[x]Bug Fix
[ ]New Feature

Benchmark results:

```
❯ node benchmark/perfTest3.js
validation : 15075.528720836128 requests/second
xml to json : 13612.412033517821 requests/second
xml to json + json string : 12826.365584690317 requests/second
xml to json string : 2125.6406392103368 requests/second
xml2js  : 5101.253171761319 requests/second
```

### DIFF
--- a/spec/validator_spec.js
+++ b/spec/validator_spec.js
@@ -361,13 +361,13 @@ describe("XMLParser", function() {
         const result = validator.validate(xmlData).err;
         expect(result).toEqual(expected);
     });
-    
+
     it("should validate XML PIs", function () {
         var xmlData = '<?xml version="1.0"?>'
         +'<?mso-contentType?>'
         +'<h1></h1>'
         +'<?mso-contentType something="val"?>';
-        
+
         var result = validator.validate(xmlData);
         expect(result).toBe(true);
     });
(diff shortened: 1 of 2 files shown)


brainsiq:
I also noticed that the package-lock.json version is out of sync with the package.json, so it changed when i npm installed. I did not commit this change.

```
diff --git a/package-lock.json b/package-lock.json
index f3de5d5..15b0485 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "fast-xml-parser",
-  "version": "3.12.5",
+  "version": "3.12.8",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
```

amitguptagwl:
Thanks @brainsiq  for the PR. I'll update package-lock.json

PR #131, opened 2018-12-11
fix for localeRange option in validator

damianed:
# Purpose / Goal
There was an error where the localeRange option of the validate function only matched the start of the attribute name

# Type
Please mention the type of PR

* [x]Bug Fix
* [ ]New Feature

### DIFF
--- a/src/validator.js
+++ b/src/validator.js
@@ -22,7 +22,7 @@ exports.validate = function(xmlData, options) {
     if (xmlData[0] === "\ufeff") {  // check for byte order mark (BOM)
       xmlData = xmlData.substr(1);
     }
-    const regxAttrName = new RegExp("^[_w][\\w\\-.:]*$".replace("_w", "_" + options.localeRange));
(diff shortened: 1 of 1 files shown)


amitguptagwl:
@damianed Thanks for your PR. Can you please share the error you're facing? I would also like to get more detail like input XML etc. Probably, raising an issue and discussing there will be more appropriate.

amitguptagwl:
@damianed any update?

damianed:
Sorry for not responding, I kept postponing it and then I forgot.
if i remember correctly, I was trying to use the library with an xml file with certain Spanish characters (e.g ñ, á, é etc) and I could not do it because it would throw an error. So it tried to fix that with this pull request. 
But I don't think it even worked (part of the reason I kept postponing this response is that I made a pull request without testing it first and I felt ashamed :( ).
Sorry, I been trying to contribute to open source  for a long time and when I find something I that I can contribute with I get over excited and rush my self.

amitguptagwl:
@damianed this is perfectly fine. If you really looking for some small issues to contribute, you can find some here: https://github.com/socialat/team

Feel free to contact back

PR #135, opened 2019-01-13
options.arrayMode support

iduchy:
the arrayMode option enebles all child elements to be in an array. It improves queries in NoSQL

### DIFF
--- a/src/node2json.js
+++ b/src/node2json.js
@@ -21,7 +21,7 @@ const convertToJson =function(node, options) {
     const keys = Object.keys(node.child);
     for (let index = 0; index < keys.length; index++) {
         var tagname = keys[index];
-        if (node.child[tagname] && node.child[tagname].length > 1) {
+        if (options.arrayMode || node.child[tagname] && node.child[tagname].length > 1) {
             jObj[tagname] = [];
             for (var tag in node.child[tagname]) {
                 jObj[tagname].push( convertToJson(node.child[tagname][tag], options) );
--- a/src/node2json.js
+++ b/src/node2json.js
@@ -21,7 +21,7 @@ const convertToJson =function(node, options) {
     const keys = Object.keys(node.child);
     for (let index = 0; index < keys.length; index++) {
         var tagname = keys[index];
-        if (options.arrayMode || node.child[tagname] && node.child[tagname].length > 1) {
+        if (node.child[tagname] && node.child[tagname].length > 1) {
             jObj[tagname] = [];
             for (var tag in node.child[tagname]) {
                 jObj[tagname].push( convertToJson(node.child[tagname][tag], options) );
--- a/src/node2json.js
+++ b/src/node2json.js
@@ -26,6 +26,9 @@ const convertToJson =function(node, options) {
             for (var tag in node.child[tagname]) {
                 jObj[tagname].push( convertToJson(node.child[tagname][tag], options) );
             }
+        }
+        else if (options.arrayMode && node.child[tagname]){
+            jObj[tagname] = [convertToJson(node.child[tagname][0], options)];
         } else {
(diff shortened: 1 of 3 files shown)


amitguptagwl:
Thanks for your PR. But this feature requires more discussion. Eg what would be the expected output for different conditions and tests to ensure that we achieve all targeted features. This feature was previously implemented and removed due to lack of clarity. Please search for closed issues.

amitguptagwl:
What would be the expected output for this?
```xml
<root>
    <item id="a">
        <sub>1</sub>
        <sub>2</sub>
    </item>
    <item id="b">
        <sub>3</sub>
    </item>
</root>
```

Can you please test your changes against [sample.xml](https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/spec/assets/sample.xml)? And please add some unit tests to validate your expectations.

iduchy:
i've added a test for the sample and removed the fature from node2json_str.js because it fails in your branch too:
```
parser = require('fast-xml-parser');
const xml = `
<root>
    <item id="a">
        <sub>1</sub>
        <sub>2</sub>
    </item>
    <item id="b">
        <sub>3</sub>
    </item>
</root>`

const options = {
    ignoreAttributes : false,
    ignoreNameSpace : false,
    allowBooleanAttributes : true,
    parseNodeValue : true,
    parseAttributeValue : true,
    arrayMode: true
};
jsonStringFromXml =parser.convertToJsonString(xml, options)
```
**output:**
```
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at _cToJsonStr (fast-xml-parser\src\node2json_str.js:19:25)
    at Object.convertToJsonString (fast-xml-parser\src\node2json_str.js:12:12)
```

amitguptagwl:
1. Can you plz rename your test file and append `_spec` in last? (just like other test files)
2. Can you plz add the tests using jasmine format? Here is the sample

```js
describe("XMLParser", function() {
    it("should parse attributes with valid names", function() {
        const xmlData = `<some>valid XML data</some>`;
        const expected = {
            "expected": "js object"
        };

        let result = parser.parse(xmlData, {
            attributeNamePrefix: "",
            ignoreAttributes:    false,
            parseAttributeValue: true,
            arrayMode: true
        });

        //console.log(JSON.stringify(result,null,4));
        expect(result).toEqual(expected);

        result = validator.validate(xmlData);
        expect(result).toBe(true);
   });
});
```

With the test file you've created what I can understand is;

input
```xml
<root foo="bar">
    <item id="a">
        <sub>1</sub>
        <sub>2</sub>
    </item>
    <item id="b">
        <sub>3</sub>
    </item>
    <item2 id="b">
        <sub>3</sub>
    </item2>
</root>
```
output
```js
{
    root:[{ 
        "@_foo": "bar",
        item:[ 
            { '@_id': 'a', sub: [ 1, 2 ] }, 
            { '@_id': 'b', sub: [ 3 ] } 
        ] ,
        item2: [{
            '@_id': 'b', sub: [ 3 ] 
        }]
    }] 
}
```
* root element would not be bound in an array.
* an array will have single object which will be having properties for attributs, text Data, cdata, and other tags
* if there is a leaf node, it will be having an array of values
* no detail about self closing tag?


Can you please generate the output against sample.xml mentioned in last comment? Because it has most of the scenarios.

iduchy:
ok, it is more complicated than i expexted. I'm going to close the pull request and open an issue soon

amitguptagwl:
As I previously suggested, there are closed issues on the same topic. You can take their reference too to understand more scenarios. :)

chovy:
Is there any progress? I too would like a "forceArray: true" for child elements.

amitguptagwl:
@chovy we need to finalize the expected result for each scenario. As that is not final yet, there is no progress.

PR #152, opened 2019-03-29
fixed attribute name for nested node

yumamurakami:
# Purpose / Goal

I tried j2x parser.
It found nested node is not been adapted option name `attrNodeName`.

I fixed this problem.
Please check it.

Thank you.

# Type
Please mention the type of PR
* [x]Bug Fix
* [ ]New Feature


amitguptagwl:
Many thanks for your PR. Can you please add a test for the same?

Issue #68, opened 2018-03-29
parse option ArrayMode not working


bb:
### Checklist
Please fill below checklist

* [x] Are you running the latest version? 3.3.6
* [x] Have you include sample input? 
* [x] Have you include actual output?
* [x] Have you include expected output?
* [x] Have you checked if you are using correct configuration?
* [x] Did you try online [tool](https://naturalintelligence.github.io/fast-xml-parser/)?
* [x] Did you [bookmarked](https://github.com/NaturalIntelligence/fast-xml-parser/stargazers) the repository for further updates? ;)

### Input data
    <x><y><z>z1</z><z>z2</z></y><w>w1</w></x>

### Code
```js
import * as FXP from "fast-xml-parser";
import { parseString } from "xml2js";

const xml = "<x><y><z>z1</z><z>z2</z></y><w>w1</w></x>";
parseString(xml, { explicitArray: false }, (err, res) => console.log("xml2js      ", JSON.stringify(res)));
parseString(xml, { explicitArray: true }, (err, res) => console.log("xml2js array", JSON.stringify(res)));
console.log("FXP         ", JSON.stringify(FXP.parse(xml)));
console.log("FXP array   ", JSON.stringify(FXP.parse(xml, { arrayMode: true })));
```

### Output data
    xml2js       {"x":{"y":{"z":["z1","z2"]},"w":"w1"}}
    xml2js array {"x":{"y":[{"z":["z1","z2"]}],"w":["w1"]}}
    FXP          {"x":{"y":{"z":["z1","z2"]},"w":"w1"}}
    FXP array    {"x":{"y":{"z":["z1","z2"]},"w":"w1"}}

### expected data
    xml2js       {"x":{"y":{"z":["z1","z2"]},"w":"w1"}}
    xml2js array {"x":{"y":[{"z":["z1","z2"]}],"w":["w1"]}}
    FXP          {"x":{"y":{"z":["z1","z2"]},"w":"w1"}}
    FXP array    {"x":{"y":[{"z":["z1","z2"]}],"w":["w1"]}}

bb:
BTW, I didn't see arrayMode in the docs, but in the source code, e.g https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/src/x2j.js#L22

If I may add a feature request here: It would be great to have the arrayMode configurable, e.g. per tag name:

 ```js
[...]

PR #160, opened 2019-04-30
Implement a way of parsing xml to closely resemble that of xml.

jinq102030:
# Purpose / Goal
Be able to parse each xml node in the form of { nodeName, attr, val, namespace, children}.
This makes it easier to process JSON from xml because there is not need to constantly check if a child node is an array or non-array.

This feature only applies when you enable resembleXml option (default is false).

### DIFF
--- a/spec/resembleXml_spec.js
+++ b/spec/resembleXml_spec.js
@@ -1,0 +1,101 @@
+"use strict";
+
+const parser = require("../src/parser");
+const validator = require("../src/validator");
+const he = require("he");
+
+describe("XMLParser", function() {
+    it("1. support parsing from xml to json with resembleXml mode", function() {
+        const xmlData = `<issue i1="i1v"><title t1="t1v">test 1</title></issue>`;
+        const expected = {
+          "nodeName": "issue",
+          "attr": {
+            "i1": "i1v"
+          },
+          "children": [
+            {
+              "nodeName": "title",
+              "attr": {
+                "t1": "t1v"
+              },
+              "val": "test 1"
+            }
+          ]
+        };
+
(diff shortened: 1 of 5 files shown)


amitguptagwl:
Thanks for the PR. Can you plz add necessary tests? So I can understand your thoughts better.

amitguptagwl:
Thanks for adding the tests to cover few scenarios. We received a feature request for some part of this implementation few days back, and we ended with the discussion that a separate schema builder should be written.

So, here, we want to understand the need and need to discuss the output structure before merging the changes. Where did you get this idea from? your personal project/need or you found the same solution somewhere else in other parsers etc.

jinq102030:
Glad someone else also has such a request. In our development,  we felt that it's very cumbersome to have to deal with aggregation: 
A xml node has 5 children, 3 of them has the same nodeName (A) and aggregated, the other two have unique nodeNames (B, C).  So the object for A is an array of 3 element, the objects for B or C are not array.   

Our code has to do lots of check to see if the object for a given nodeName is an array or not,  it created so much if else conditions and it's becomes hard to manage.

amitguptagwl:
So the main issue I see here is to have the elements in array. I'll suggest you go through following issues and PR;

https://github.com/NaturalIntelligence/fast-xml-parser/issues/68
https://github.com/NaturalIntelligence/fast-xml-parser/pull/135

jinq102030:
Thanks for the suggestion, will check it out.  Sorry for the late reply.

amitguptagwl:
As we have published arrayMode now. I'm closing this PR. Thanks for your effort.

PR #163, opened 2019-05-20
Fix: Remove polyfill

### LINKED ISSUE #162: Does not work in IE11 Web Worker due to use of window
Due to [this code](https://github.com/NaturalIntelligence/fast-xml-parser/blob/65c39beae792e643cd00a6dd866b31178afcf16b/src/xmlstr2xmlnode.js#L13-L19):
```
//polyfill
if (!Number.parseInt && window.parseInt) {
  Number.parseInt = window.parseInt;
}
if (!Number.parseFloat && window.parseFloat) {
  Number.parseFloat = window.parseFloat;
}
```
Would it be better to leave polyfills out completely and direct people to use polyfill.io?

timfish:
Removes the need to polyfill `Number.parseInt|parseFloat` by simply using the top level `parseInt|parseFloat`.

Fixes #162

### DIFF
--- a/src/xmlstr2xmlnode.js
+++ b/src/xmlstr2xmlnode.js
@@ -10,14 +10,6 @@ let regx =
 //const tagsRegx = new RegExp("<(\\/?[\\w:\\-\._]+)([^>]*)>(\\s*"+cdataRegx+")*([^<]+)?","g");
 //const tagsRegx = new RegExp("<(\\/?)((\\w*:)?([\\w:\\-\._]+))([^>]*)>([^<]*)("+cdataRegx+"([^<]*))*([^<]+)?","g");
 
-//polyfill
-if (!Number.parseInt && window.parseInt) {
-  Number.parseInt = window.parseInt;
-}
-if (!Number.parseFloat && window.parseFloat) {
-  Number.parseFloat = window.parseFloat;
-}
-
 const defaultOptions = {
   attributeNamePrefix: '@_',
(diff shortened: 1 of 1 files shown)


amitguptagwl:
Sorry I didn't understand that how are you ensuring it to work on IE?

timfish:
I changed `Number.parseFloat` to simply use `parseFloat`. `parseFloat` exists even on IE6.

amitguptagwl:
We knowingly placed them to fix an error when we found them not present in some browser. You can probably check in closed issues for more detail.

timfish:
I don't think you've understood my changes correctly.

`Number.parseInt` does not exist in older versions of IE.

To fix this, it looks like a polyfill [was added](https://github.com/NaturalIntelligence/fast-xml-parser/commit/1c4ea5c1e26df660a70321327b53564aa13b8e17) which copies `parseInt` from global window `Number.parseInt = window.parseInt`.

But the above polyfill has mistakes in it. #162 highlights this. `window` does not exist in all JavaScript contexts.

Rather than use `Number.parseInt` you should just be using `parseInt` on its own which works in every browser released in the lat 10 years. 

This PR fixed the issue with `window` and still works in every browser.

amitguptagwl:
I'm fine with having in this case

```js
if (!Number.parseInt) {
  Number.parseInt = parseInt;
}
if (!Number.parseFloat) {
  Number.parseFloat = parseFloat;
}
```

PR #164, opened 2019-05-20
fix validation for tag on multiple lines

francescobarbera:
When a tag is split into multiple lines, validation fails.
For example
```
<name
attribute1="attribute1"
attribute2="attribute2"
/>
```
is considered invalid.

To solve this issue, I add a check also for new line characters.


amitguptagwl:
Thanks for your changes.

Issue #124, opened 2018-11-08
location of validation error


jzou2000:
Hi
I think it would be better to give line/column of error when the validation fails. Those information are very useful when the xml file is large. xml2js (and most validators) gives those info.

Currently the err object contains only code and msg. I checked the source code of validate and it seems it is not a config issue.

PS. is it possible to validate while parsing? If I validate then parse, the speed reduce half, if I don't validate, the parser parses silently but gives the wrong result.

Thanks.

amitguptagwl:
If we combine, parsing speed will definitely reduce. As in most of the cases, we need not validate the trusted resources it is important to keep them separate.

Line number can be calculated but in some cases, it can be inaccurate. Let be mark it as a feature request so we can do something helpful. However, I can't commit any timeline to get it done.

mubaidr:
An `validate` option can be added to the method, with defaults set to `false`. For users library will be fast and if someone wants to validate, he/she can set it to true.

amitguptagwl:
@mubaidr though it's a good suggestion, @jzou2000 has the concern about overall throughput when validation and parsing are combined.

amitguptagwl:
@mubaidr , @jzou2000 we have changed the method signature to optionally choose if validation need to be done.

NorbertHuethmayr:
Any news on the initial request to include the line number when validation runs into an error?

amitguptagwl:
No changes done for line number in response yet. However, if someone is interested to raise the PR please visit line number 137 in validator.js

PR #198, opened 2019-11-05
Validation errors now include line numbers in the error message

NorbertHuethmayr:
# Purpose / Goal
Whenever the validator encounters an error, the line number is not part of the error message. 

This issue has been described in issue #124

# Implementation Detail
The line number is obtained not be counting the lines as we go through the code, rather by splitting the lines at line break symbols and counting the resulting lines. The performance of this library is not affected as the code executes in an error case only. It also blends in with the existing code much nicer.

# Type
Please mention the type of PR
* [ ]Bug Fix
* [x ]New Feature


amitguptagwl:
Thanks for raising this PR. Can you please resolve the conflict? I'll check this PR over this weekend.

NorbertHuethmayr:
> Thanks for raising this PR. Can you please resolve the conflict? I'll check this PR over this weekend.

Done -  Test passing again.

NorbertHuethmayr:
Hi Amit, I found some cases where the calculation is not correct. Please wait for me to fix those problems. I will let you know when you can merge.

Best regards
Norbert

NorbertHuethmayr:
@amitguptagwl 

I've added more tests to verify that the line numbers are good even in case there are multi-line attributes. I had to choose a different approach than at the beginning which is why it took a bit longer.

Let me know if you need more, from my point of view you can go ahead and merge.

Regards
Norbert

NorbertHuethmayr:
@amitguptagwl , 

sorry for the mess. I'm too unfamiliar with Git and the tools around it :(
I had to repair a broken merge which is always kinda messy. Please have a look at this PR so I can wrap this up.

Best regards
Norbert

amitguptagwl:
Thanks for your effort, Norbert. Looks like I'm not gonna be super busy for the next 1 week. But I'll still try to review this PR ASAP.

NorbertHuethmayr:
Ok cool,

No big deal in fact, in the meantime I will use my forked version.
The other empty email was a mistake… sorry for that.

Best regards
Norbert


On 13.11.2019, at 19:56, Amit K Gupta <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks for your effort, Norbert. Looks like I'm not gonna be super busy for the next 1 week. But I'll still try to review this PR ASAP.

src/validator.js
```diff
@@ -75,12 +75,23 @@ exports.validate = function(xmlData, options) {
           i--;
         }
         if (!validateTagName(tagName, regxTagName)) {
-          return {err: {code: 'InvalidTag', msg: 'Tag ' + tagName + ' is an invalid name.'}};
+          return {
```
amitguptagwl: Though it is not related to your PR, can you please add this code?

```js
 if (!validateTagName(tagName, regxTagName)) {
          let msg;
          if(tagName.trim().length === 0) {
            msg = "There is an unnecessary space between tag name and backward slash '</ ..' ";
          }else{
            msg = `Tag ${tagName} is an invalid name.`;
          }
          return {
            err: {
              code: 'InvalidTag', msg: msg,
              line: getLineNumberForPosition(xmlData, i),
            },
          };
        }
```

This will elaborate the error when there is extra space between tagname and closing slash.

src/validator.js
```diff
@@ -287,27 +344,25 @@ function validateAttributeString(attrStr, options, regxAttrName) {
   const attrNames = {};
 
   for (let i = 0; i < matches.length; i++) {
-    //console.log(matches[i]);
-
     if (matches[i][1].length === 0) {
       //nospace before attribute name: a="sd"b="saf"
-      return {err: {code: 'InvalidAttr', msg: 'attribute ' + matches[i][2] + ' has no space in starting.'}};
+      return {err: {code: 'InvalidAttr', msg: 'attribute ' + matches[i][2] + ' has no space in starting.', position: getPositionFromMatch(attrStr, matches[i][0])}};
```
amitguptagwl: Though it is existing, can you please surround the attribute name in quotes to make it more readable on CLI?

```js
msg: `attribute '${matches[i][2]}' has no space in starting.`
```

src/validator.js
```diff
@@ -93,29 +104,59 @@ exports.validate = function(xmlData, options) {
...
+                },
               };
             }
           }
         } else {
           const isValid = validateAttributeString(attrStr, options, regxAttrName);
           if (isValid !== true) {
-            return isValid;
+            //the result from the nested function returns the position of the error within the attribute
+            //in order to get the 'true' error line, we need to add the calculate the position where the attribute begins (i - attrStr.length) and then add the position within the attribute
```
amitguptagwl: I believe you want to write "we need to calculate" instead of "we need to add the calculate"

src/validator.js
```diff
@@ -93,29 +104,59 @@ exports.validate = function(xmlData, options) {
             tagFound = true;
             //continue; //text may presents after self closing tag
           } else {
-            return isValid;
+            //the result from the nested function returns the position of the error within the attribute
+            //in order to get the 'true' error line, we need to add the calculate the position where the attribute begins (i - attrStr.length) and then add the position within the attribute
```
amitguptagwl: I believe you want to write "we need to calculate" instead of "we need to add the calculate"

amitguptagwl:
@NorbertHuethmayr Your changes are perfect. There are a few additional changes not related to your PR though. Please have a look.

NorbertHuethmayr:
Hi Amit,

will do, cant get to it until Monday evening though :(
I'll text you when its done.

Best regards
Norbert

amitguptagwl:
As the code climate showing 19 observations; Can you plz do the following?

1. Create a separate function to create an error object, and call with code, msg, and line number.

This will fix the maximum issues.

NorbertHuethmayr:
Sure, good idea

________________________________

src/validator.js
NorbertHuethmayr: Hi Amit,

I do agree that this should be fixed, but I also don't want to mix in to mix too many unrelated changes into one PR. 
This is probably worth another issue.

Best regards
Norbert

src/validator.js
NorbertHuethmayr: Hi Amit,

same as below - this is unrelated and need additional tests and so forth.
We should probably create an issue and then merge that one separately.

Best regards
Norbert

src/validator.js
amitguptagwl: agree. Will do it before release.

PR #201, opened 2019-11-07
Fix validation of self-closing tag followed by text

orgads:
If there is no space before the closing tag, and the tag is followed by text, validation failed:

```
<main><empty/>text</main>
```

# Purpose / Goal

# Type
Please mention the type of PR

* [X]Bug Fix
* [ ]New Feature

### DIFF
--- a/spec/validator_spec.js
+++ b/spec/validator_spec.js
@@ -50,19 +50,41 @@ describe("XMLParser", function() {
     });
 
     it("should validate self closing tags", function() {
-        const xmlData = "<rootNode><validtag1  /><validtag2/><validtag3  with='attrib'/></rootNode>";
+        const xmlData = "<rootNode><validtag1  /><validtag2/><validtag3  with='attrib'/><validtag4 />text<validtag5/>text</rootNode>";
         const result = validator.validate(xmlData);
         expect(result).toBe(true);
     });
 
-    it("should not validate self closing tags", function() {
-        const xmlData = "<rootNode><validtag1/><invalid tag/><validtag3  with='attrib'/></rootNode>";
-        const expected = {code: "InvalidAttr", msg: "boolean attribute tag is not allowed."};
+    it("should not consider these as self closing tags", function() {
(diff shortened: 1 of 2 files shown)


amitguptagwl:
Thanks for raising this PR. LEt me go through the changes.

orgads:
Added another test. Would you prefer that on this case the error would be "Tag in/valid is an invalid name"?

spec/validator_spec.js
```diff
@@ -64,6 +64,14 @@ describe("XMLParser", function() {
         expect(result).toEqual(expected);
     });
 
+    it("should not consider this as self-closing tag", function() {
+        const xmlData = "<rootNode><in/valid></rootNode>";
+        const expected = {code: "InvalidAttr", msg: "attribute /valid has no space in starting."};
```
orgads: Is this expected to be InvalidTag, and "Tag in/valid is an invalid name"? Or are you ok with this error?

orgads:
I changed this to behave just like `<foo />`. When I added tests I noticed some more bugs, so I opened separate PRs for them. If you want me to reorder the commits I can do it. Tests will need to be adapted on this case.

spec/validator_spec.js
```diff
@@ -50,19 +50,41 @@ describe("XMLParser", function() {
...
         expect(result).toEqual(expected);
+
+        xmlData = "<rootNode><notSelfClosing/ ></rootNode>";
+        expected = {code: "InvalidAttr", msg: "attribute / has no space in starting."};
+
+        result = validator.validate(xmlData).err;
+        expect(result).toEqual(expected);
+    });
+
+    /*
```
orgads: This test is commented out because it doesn't pass due to another bug. This is addressed in https://github.com/NaturalIntelligence/fast-xml-parser/pull/205

orgads:
This fix is minor, so I reordered the commits to have it first. The other unicode PR now depends on this change.

amitguptagwl:
By inspiring your this fix, I've just pushed a commit. Please have a look and check if it solves the purpose of both PR. If not you can revise your PR.

amitguptagwl:
@orgads Do we need this PR anymore?

PR #205, opened 2019-11-08
Validator: Fix name regexp

orgads:
# Purpose / Goal
Validator: Fix name regexp

The previous implementation only allowed the "localeRange" for the first character, and it did not validate the following characters at all, because the regexp did not include $, so if the first character
matched, the regexp would match.

Replace the regexp with one that matches the [XML spec](https://www.w3.org/TR/xml/#NT-NameStartChar) (except [\u10000-\uEFFFF] which matches digits for some reason...) and match *all* the characters.

Remove the localeRange option, which is no longer needed

# Type
Please mention the type of PR

* [X]Bug Fix
* [ ]New Feature

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -1,5 +1,10 @@
 'use strict';
 
+const nameStartChar = ':A-Za-z_\\u00C0-\\u00D6\\u00D8-\\u00F6\\u00F8-\\u02FF\\u0370-\\u037D\\u037F-\\u1FFF\\u200C-\\u200D\\u2070-\\u218F\\u2C00-\\u2FEF\\u3001-\\uD7FF\\uF900-\\uFDCF\\uFDF0-\\uFFFD';
+const nameChar = nameStartChar + '\\-.\\d\\u00B7\\u0300-\\u036F\\u203F-\\u2040';
+const nameRegexp = '[' + nameStartChar + '][' + nameChar + ']*'
+const regexName = new RegExp('^' + nameRegexp + '$');
+
 const getAllMatches = function(string, regex) {
   const matches = [];
   let match = regex.exec(string);
@@ -15,13 +20,9 @@ const getAllMatches = function(string, regex) {
   return matches;
 };
 
-const doesMatch = function(string, regex) {
-  const match = regex.exec(string);
+const isName = function(string) {
+  const match = regexName.exec(string);
   return !(match === null || typeof match === 'undefined');
-};
-
-const doesNotMatch = function(string, regex) {
-  return !doesMatch(string, regex);
 };
 
 exports.isExist = function(v) {
@@ -81,6 +82,6 @@ exports.buildOptions = function(options, defaultOptions, props) {
   return newOptions;
 };
 
-exports.doesMatch = doesMatch;
-exports.doesNotMatch = doesNotMatch;
+exports.isName = isName;
 exports.getAllMatches = getAllMatches;
+exports.nameRegexp = nameRegexp;
--- a/src/validator.js
+++ b/src/validator.js
@@ -4,10 +4,9 @@ const util = require('./util');
 
 const defaultOptions = {
   allowBooleanAttributes: false, //A tag can have attributes without any value
-  localeRange: 'a-zA-Z',
 };
 
-const props = ['allowBooleanAttributes', 'localeRange'];
+const props = ['allowBooleanAttributes'];
 
 //const tagsPattern = new RegExp("<\\/?([\\w:\\-_\.]+)\\s*\/?>","g");
 exports.validate = function (xmlData, options) {
@@ -16,12 +15,6 @@ exports.validate = function (xmlData, options) {
   //xmlData = xmlData.replace(/(\r\n|\n|\r)/gm,"");//make it single line
   //xmlData = xmlData.replace(/(^\s*<\?xml.*?\?>)/g,"");//Remove XML starting tag
   //xmlData = xmlData.replace(/(<!DOCTYPE[\s\w\"\.\/\-\:]+(\[.*\])*\s*>)/g,"");//Remove DOCTYPE
-  const localRangeRegex = new RegExp(`[${options.localeRange}]`);
-
-  if (localRangeRegex.test("<#$'\"\\\/:0")) {
-    return getErrorObject('InvalidOptions', 'Invalid localeRange', 1);
-  }
-
   const tags = [];
   let tagFound = false;
 
@@ -32,8 +25,6 @@ exports.validate = function (xmlData, options) {
     // check for byte order mark (BOM)
     xmlData = xmlData.substr(1);
   }
-  const regxAttrName = new RegExp(`^[${options.localeRange}_][${options.localeRange}0-9\\-\\.:]*$`);
-  const regxTagName = new RegExp(`^([${options.localeRange}_])[${options.localeRange}0-9\\.\\-_:]*$`);
   for (let i = 0; i < xmlData.length; i++) {
     if (xmlData[i] === '<') {
       //starting of tag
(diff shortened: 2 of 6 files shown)


amitguptagwl:
`localeRange` option is aligned with XML to JSON parser so that if someone parse the js object back to XML, he should get the same output. Hence, we can't remove it. Please let me know what are the bugs that you see. Currently, if `localrange` is specified then only that certain characters should be allowed.

amitguptagwl:
Please check my latest commit

orgads:
It solves the issues I found.

I'm still not sure why you need localeRange. Both XML and JSON support unicode characters in names. Why don't you just allow any supported unicode the spec defines? It can be converted both ways.

amitguptagwl:
Yes, you're right. We can allow that. But in this case we'll have to remove `localerange` from parser and validator both.

amitguptagwl:
You're so quick. Thanks. Let me check.

As this will be a change in the API, I'll have to publish it with minor version change. I'll try to do that ASAP.

orgads:
Please also update the Readme and release notes.

amitguptagwl:
I'll try to review the changes tomorrow morning :) Thanks for your effort

src/util.js
```diff
@@ -81,6 +81,10 @@ exports.buildOptions = function(options, defaultOptions, props) {
   return newOptions;
 };
 
+const nameStartChar = ':A-Za-z_\\u00C0-\\u00D6\\u00D8-\\u00F6\\u00F8-\\u02FF\\u0370-\\u037D\\u037F-\\u1FFF\\u200C-\\u200D\\u2070-\\u218F\\u2C00-\\u2FEF\\u3001-\\uD7FF\\uF900-\\uFDCF\\uFDF0-\\uFFFD';
+const nameChar = nameStartChar + '\\-.\\d\\u00B7\\u0300-\\u036F\\u203F-\\u2040';
+exports.nameRegexp = '[' + nameStartChar + '][' + nameChar + ']*'
+
```
amitguptagwl: Can it be a good idea of defining the range of letters that should not be included in the tag name? Because any other Unicode character should be allowed. Moreover, there can be more ranges I believe.

src/validator.js
```diff
@@ -316,17 +308,17 @@ function validateAttributeString(attrStr, options, regxAttrName) {
 
 // const validAttrRegxp = /^[_a-zA-Z][\w\-.:]*$/;
 
-function validateAttrName(attrName, regxAttrName) {
-  // const validAttrRegxp = new RegExp(regxAttrName);
-  return util.doesMatch(attrName, regxAttrName);
+function validateAttrName(attrName, regexName) {
```
amitguptagwl: Since `regexName` is gonna be constant, do we need to pass it?

src/validator.js
```diff
@@ -316,17 +308,17 @@ function validateAttributeString(attrStr, options, regxAttrName) {
...
+  // const validAttrRegxp = new RegExp(regexName);
+  return util.doesMatch(attrName, regexName);
 }
 
 //const startsWithXML = new RegExp("^[Xx][Mm][Ll]");
 //  startsWith = /^([a-zA-Z]|_)[\w.\-_:]*/;
 
-function validateTagName(tagname, regxTagName) {
+function validateTagName(tagname, regexName) {
```
amitguptagwl: Since `regexName` is gonna be constant, do we need to pass it?

amitguptagwl:
@orgads are you still on this?

src/util.js
orgads: Which other ranges? I took the values from xml spec.

src/validator.js
orgads: Will do

src/util.js
amitguptagwl: It was my doubt. But I tested this for randomly for 15 languages and they're covered. So it seems fine. Now, my concern is performance. Have you checked the impact on performance when by default all the ranges are applicable?

orgads: Looks like this change *improves* performance. I ran with and without the change - each commit twice, and picked the better result.

Before:
```
validation : 12234.247180941054 requests/second
xml to json : 10602.1257025917 requests/second
xml to json + json string : 7859.281152576787 requests/second
xml to json string : 1747.9156403661627 requests/second
xml2js  : 2985.894748422982 requests/second
```

After:
```
validation : 13801.418604604587 requests/second
xml to json : 13016.920442592424 requests/second
xml to json + json string : 10186.32443980399 requests/second
xml to json string : 1807.3535925877125 requests/second
xml2js  : 3310.3600184283914 requests/second
```

orgads:
I removed localeRange from README.md and index.html, but did not remove it from parser.d.ts in order to not break existing code that uses it.

src/util.js
amitguptagwl: Thanks for the update. I'll probably test on my machine too.

amitguptagwl:
Changes in `parser.d.ts` should not break anything. It'll just give syntax errors to the existing users. Anyhow, we can't release these changes under the minor version. So the changes to `parser.d.ts` would be fine.

orgads:
The failed checks are false positives.

amitguptagwl:
Sorry for the delay. But I'm unable to take any action for next few weeks. Will update you. And thanks for your effort.

amitguptagwl:
I have cross verified for the performance. It looks fine. Let me check for other aspects. I'll update you if there is any issue.

src/xmlstr2xmlnode.js
```diff
@@ -4,8 +4,9 @@ const util = require('./util');
 const buildOptions = require('./util').buildOptions;
 const xmlNode = require('./xmlNode');
 const TagType = {OPENING: 1, CLOSING: 2, SELF: 3, CDATA: 4};
-let regx =
-  '<((!\\[CDATA\\[([\\s\\S]*?)(]]>))|(([\\w:\\-._]*:)?([\\w:\\-._]+))([^>]*)>|((\\/)(([\\w:\\-._]*:)?([\\w:\\-._]+))\\s*>))([^<]*)';
+const regx =
+  '<((!\\[CDATA\\[([\\s\\S]*?)(]]>))|((NAME:)?(NAME))([^>]*)>|((\\/)(NAME)\\s*>))([^<]*)'
```
amitguptagwl: Can you please recheck this regex?

Expected regex is : `<((!\\[CDATA\\[([\\s\\S]*?)(]]>))|((NAME:)?(NAME))([^>]*)>|((\\/)((NAME:)?(NAME))\\s*>))([^<]*)`

orgads: This is redundant. The previous one can also be removed. Colon is a valid character everywhere in the name, so there's no reason to have it explicitly stated in the regex, especially when it is not mandatory [i.e. `(NAME:)?`].

amitguptagwl: `:` can not be repeated in name. Hence, taking in NAME regex but not in 2nd regex will not be correct. However, as the validator is separate functionality we can have it in the way you did it. But in this case, I see, there is no point of having in first part as well. right?

`'<((!\\[CDATA\\[([\\s\\S]*?)(]]>))|(NAME)([^>]*)>|((\\/)(NAME)\\s*>))([^<]*)'`

Considering this, we can even move this logic to the validator `const nameRegexp = '[' + nameStartChar + '][' + nameChar + ']*'`, instead of keeping it in parser.

orgads: From the spec:

> Note:
> 
> The Namespaces in XML Recommendation [XML Names] assigns a meaning to names containing colon characters. Therefore, authors should not use the colon in XML names except for namespace purposes, but **XML processors must accept the colon as a name character**.
> 
> ...
> 
> Document authors are encouraged to use names which are meaningful words or combinations of words in natural languages, and to avoid symbolic or white space characters in names. **Note that COLON, HYPHEN-MINUS, FULL STOP (period), LOW LINE (underscore), and MIDDLE DOT are explicitly permitted**.

Regarding the first appearance in the regex, like I said, it is needed for the ignoreNamespace option.

amitguptagwl: Agree.. I believe this needs to be handled properly by comparing the namespace name instead of its syntax only (probably in version FXP v4).

Issue #207, opened 2019-11-12
Multiple root elements validate successfully


NorbertHuethmayr:
### Checklist

* [x] Have you asked your question on Stackoverflow or similar forum?
* [x] Are you running the latest version?
* [ ] Have you included sample input, output, error, and expected output? 
* [x] Have you checked if you are using correct configuration?
* [x] Did you try online [tool](https://naturalintelligence.github.io/fast-xml-parser/)?

 
### Input
XML structures must always have a single root element. If there are multiple, it is not valid. The validator class does not detect this problem.

### Expected data
I would expect a descriptive error message when an XML with multiple root-nodes is validated.

**Would you like to work on this issue?**

- [x] Yes
- [ ] No

[Bookmark](https://github.com/NaturalIntelligence/fast-xml-parser/stargazers) this repository for further updates.

NorbertHuethmayr:
Hi Amit,

if you give me a hint on where to start, I could get going.

amitguptagwl:
Probably, you can maintain a stack. push when a tag is found, pop when matching closing tag is found. If the size of the stack reaches to 0 and you still have more tags, it means there are more than 1 root element.

PR #210, opened 2019-11-23
Single root element only

NorbertHuethmayr:
# Purpose / Goal
This PR fixes issue #207 

# Type
Please mention the type of PR
* [x] Bug Fix
* [ ] New Feature


amitguptagwl:
Will this code work for repeated nodes? 
```xml
<root></root><root></root>
```

NorbertHuethmayr:
> Will this code work for repeated nodes?
> 
> ```
> <root></root><root></root>
> ```

Hi Amit,

the code you mentioned is exactly what is invalid as it does not have a single root element.
If multiple nested elements are contained within a single root element, then yes, it will work.

I also added 2 unit tests to verify just that.

Best regards
Norbert

amitguptagwl:
Your PR is testing
```xml
<xml></xml><xml2></xml2>
```
and
```xml
<test>
        <nested>
        </nested>
</test>
<xml2>
</xml2>
```
but not
```xml
<xml></xml><xml></xml>
```

Please also read my comment on the issue you raised for the same.

NorbertHuethmayr:
You are right - the code i have pushed is not working properly. I will work on a better fix....

NorbertHuethmayr:
@amitguptagwl 

I've improved the detection so that the depth of a node is tracked. If the root level is reached more than once, an error (including the line) is created.
I also had to change your "large" file as it was technically invalid (it had multiple root nodes)

Let me know what you think of this PR.

Best regards
Norbert

src/validator.js
```diff
@@ -131,10 +135,19 @@ exports.validate = function (xmlData, options) {
...
+            if(reachedRoot)
+            {
+              return getErrorObject('InvalidXml', 'Multiple possible root nodes found.', getLineNumberForPosition(xmlData, i));
+            }
 
-          //we remember the first opening tag that we find
-          if (firstOpeningTag === '') {
-            firstOpeningTag = tagName;
+            //otherwise we set a flag for the next time around
+            reachedRoot = true;
```
amitguptagwl:
```js
if (tags.length === 0) {
    reachedRoot = true;
}
```

src/validator.js
```diff
@@ -131,10 +135,19 @@ exports.validate = function (xmlData, options) {
           }
           tags.push(tagName);
```
amitguptagwl: if(reachedRoot === true){
    return getErrorObject('InvalidXml', 'Multiple possible root nodes found.', getLineNumberForPosition(xmlData, i));
}else{
    tags.push(tagName);
}

NorbertHuethmayr: Good point, will do...

amitguptagwl:
Changes look fine in the validator. I'll check other files nad merge the changes as soon as possible.

NorbertHuethmayr:
> Big file increases the of the repo and user faces the issue at the time of cloning. Hence, I was wondering if you can remove the new large file. And use some code like this in the test;
> 
> ```js
> const xmlData = fs.readFileSync(fileNamePath).toString();
> //xmlData=`<root>${xmlData.repeat(1000)}</root>`;
> ```
> 
> In this way, we can even control the size of the data on the fly. Eventually, we'll remove other big files as well from the project.

Hi Amit,

I pretty much disagree with you there. The size of a repo cant be a concern. Imagine really big projects with thousands of files....

One thing that I could do is to cut out half of the content from the copied file.
Would that be ok?

amitguptagwl:
> Hi Amit,
>
>I pretty much disagree with you there. The size of a repo cant be a concern. Imagine really big projects with thousands of files....
>
>One thing that I could do is to cut out half of the content from the copied file.
Would that be ok?

A big project is different than a project with bigger files. Large XMl that we have in our project is of 100mb. You must be having good internet speed, I don't. I sometimes need to clone the project of all the PRs to test them thoroughly which requires more time and bandwidth.

In addition, this is not that something we should compromise due to any limitations. It gives us more control.

spec/validator_spec.js
```diff
@@ -511,4 +511,33 @@ attribute2="attribute2"
...
+        </test>
+        <xml2>
+        </xml2>`;
+
+        const expected = { code: "InvalidXml", msg: 'Multiple possible root nodes found.', line: 5};
+        var result = validator.validate(xmlData).err;
+        expect(result).toEqual(expected);
+    });
+
+    it('should not validate XML documents with multiple root nodes 3', () => {
```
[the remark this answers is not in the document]
NorbertHuethmayr: Done

NorbertHuethmayr:
> ```js
> when root nodes have more nested tags
> ```

Hi Amit,

I introduced a new group and grouped the existing tests as you suggested.

NorbertHuethmayr:
> > Hi Amit,
> > I pretty much disagree with you there. The size of a repo cant be a concern. Imagine really big projects with thousands of files....
> > One thing that I could do is to cut out half of the content from the copied file.
> > Would that be ok?
> 
> A big project is different than a project with bigger files. Large XMl that we have in our project is of 100mb. You must be having good internet speed, but all the contributors might not have. I sometimes need to clone the project of all the PRs to test them thoroughly which requires more time and bandwidth.
> 
> In addition, this is not that something we should compromise due to any limitations. It gives us more control. Think it in this way, what is the advantage of having large files in a project when you can eliminate them?
> 
> Or think it to optimize the existing available assets. Eg. you can use `ptest.xml` and make it as bigger as you want in the test as per your requirement. I would be happy to hear the cons of this approach.

I've removed the 2nd very large file and the unit test that used it. It was a duplicate test anyway and didnt do much in the first place.

amitguptagwl:
Thanks for your contribution.

PR #224, opened 2020-01-16
Typo supress->suppress

caub:
# Purpose / Goal
The typo in 'supress' bugged me

This would be a breaking change


amitguptagwl:
Thanks for your PR. This PR needs to be released with a major version change. So I'm currently keeping it on hold for some time. Will update

PR #225, opened 2020-01-20
Fix issue where invalid xml could be generated...

johnsusek:
Fix issue where invalid xml could be generated from JSON without a root note.

# Purpose / Goal
This patch adds a `<root>` node to the output xml if the input JSON has no root nodes.

# Type
* [x]Bug Fix

### DIFF
--- a/spec/j2x_spec.js
+++ b/spec/j2x_spec.js
@@ -202,6 +202,17 @@ describe("XMLParser", function() {
         expect(result).toEqual(expected);
     });
 
+    it("should add root node to xml if there is none", function() {
+        const jObj = {
+            a: { "foo": "bar" },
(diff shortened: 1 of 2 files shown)


amitguptagwl:
Thanks for your effort. Can you please add a test to verify your changes?

johnsusek:
Sure, I will do that shortly. What do you think about making this an option? I am concerned about changing existing behavior because maybe people are expecting it

johnsusek:
Unit test added

amitguptagwl:
I'm just wondering if it should be based on user's choice. Because it'll silently add the root when not present. So what if `closeRoot` can be replaced with `closeRootIfNotPresent` parsing option?

PR #237, opened 2020-03-30
feat: add supressEmptyNodeWithoutAttributes option

iteixeira:
# Purpose / Goal
I needed to convert a JSON attribute with an empty value to a not self closing XML tag:

**Input:**

```
{
  "name": "",
  "address": ""
}
```

**Expected output:**

```
<name></name><address></address>
```

**Actual output:**

```
<name/><address/>
```

# Benchmark results

Results of the `node benchmark\perfTest3.js` command, as requested at **CONTRIBUTING.md**:

**Before changes:**

validation : 22506.73362810279 requests/second
xml to json : 23379.649527832113 requests/second
xml to json + json string : 20339.19984773564 requests/second
xml to json string : 3201.794682753688 requests/second
xml2js  : 5631.943439669823 requests/second

**After changes:**

validation : 22660.97418380873 requests/second
xml to json : 22143.350711227682 requests/second
xml to json + json string : 19650.172211612786 requests/second
xml to json string : 3152.241109975055 requests/second
xml2js  : 5766.8773762699975 requests/second

# Type
Please mention the type of PR

* [ ]Bug Fix
* [X]New Feature

### DIFF
--- a/spec/j2x_spec.js
+++ b/spec/j2x_spec.js
@@ -309,6 +309,27 @@ describe("XMLParser", function() {
         expect(result).toEqual(expected);
     });
 
+    it("should not supress empty node without attributes to self closing node when parsing to XML", function() {
+        const jObj = {
+            name: '',
+            address: ''
+        };
+        const parser = new Parser({
+                                      cdataTagName:     "__cdata",
+                                      attributeNamePrefix: "",
+                                      attrNodeName:     "@",
+                                      encodeHTMLchar:   true,
+                                      supressEmptyNode: true,
+                                      supressEmptyNodeWithoutAttributes: false,
+                                      tagValueProcessor: a=> { a= ''+ a; return he.encode(a, { useNamedReferences: true}) },
+                                      attrValueProcessor: a=> he.encode(a, {isAttributeValue: true, useNamedReferences: true})
+                                  });
+        const result = parser.parse(jObj);
+        //console.log(result);
+        const expected = `<name></name><address></address>`;
+        expect(result).toEqual(expected);
+    });
+
     it("should format when parsing to XML", function() {
         const jObj = {
             a: {
--- a/src/json2xml.js
+++ b/src/json2xml.js
@@ -12,6 +12,7 @@ const defaultOptions = {
   format: false,
   indentBy: '  ',
   supressEmptyNode: false,
+  supressEmptyNodeWithoutAttributes: true,
   tagValueProcessor: function(a) {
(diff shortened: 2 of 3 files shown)


amitguptagwl:
What output do you get when you set `supressEmptyNode: false` without your changes?

iteixeira:
If I set `supressEmptyNode: false`, I'll get something like:

**JSON:**
```
{
  "person": {
    "@_name: "John"
  },
  "address": ""
}
```

**XML**:
```
<person "name"="John"></person>
<address></address>
```

**But I need the XML to be like:**

```
<person "name"="John"/>
<address></address>
```

To get the result above, I have to set `supressEmptyNode: true` and `supressEmptyNodeWithoutAttributes: false`, then it won't supreess tags without attributes (`<address></address>`).

amitguptagwl:
Thanks. Now it is more clear to me. However, the option name is still confusing `supressEmptyNodeWithoutAttributes`. It may be understood as an option to suppress a node excluding its attributes. I'll have to think more on both options.

iteixeira:
Yes, thinking about it now, it's confuse. Sorry about that. But I didn't think a better name for it,

iteixeira:
Yes, thinking about it now, it's confuse. Sorry about that. But I didn't think a better name for it. What do you think if I create an `supressEmptyNodeConditional` or `supressEmptyNodeConditionalProcessor`, where you could set a function that receive the node infos as parameter (key, value and attrs), and returns a boolean value? With that, I can supress the node conditionally.

Example:

```
const j2xParser = new parser.j2xParser({
  ignoreAttributes: false,
  attributeNamePrefix: '@_',
  supressEmptyNode: true,
  supressEmptyNodeConditionalProcessor: (key, value, attrs) => attrs === '',
});
```

In that way, I think there will be no doubt about the option name meaning, and it will work for what I need.

amitguptagwl:
What do you think about `emptyNodeProcessor`? If it can decide 
1. if a node without attributes should be converted to self closing node.
2. if a node with attributes should be converted to self closing node.
3. if a node should be assigned with some default value
etc.

However, I strongly feel that if `supressEmptyNode:true` then a node with or without attributes should convert to self-closing tag instead of another option. It may confuse people.

amitguptagwl:
```xml
<car>
  <color alpha="7"/>
  <type>minivan</type>
  <registration>2020-02-03</registration>
  <capacity>7</capacity>
</car>
```

PR #268, opened 2020-05-29
fix: unexpected empty string

### LINKED ISSUE #267: empty string added in json array due to custom Array prototype

### Description

hi, guys, I met the bug of unexpected tailing empty string in array when converting xml to json today, after debug, I found [this line](https://github.com/NaturalIntelligence/fast-xml-parser/blob/057a2d31f112ba9e7bb4afa5bec5c8675f86c36e/src/node2json.js#L31)
enumerated the Array custom prototype and cause the empty string pending, maybe we should exam whether `tag` is `node.child[tagname]` own property?

### Input

just use the online tool, and add custom Array prototype member using the devtools, then the bug would show up.

### Code

```js
Array.prototype.test = 'test';
```

- [x] Yes
- [ ] No

I'm not sure it's appropriate to check with `hasOwnProperty` to check the member in every for...in loop

Shana-AE:
fix unexpected empty string caused by for...in iterating over Array custom prototype properties.
by adding hasOwnProperty check in for..in iterator.

# Purpose / Goal

check own property in for...in iterator. 

Fixes #267 

# Type

Please mention the type of PR

* [x]Bug Fix
* [ ]New Feature

# Benchmark

validation : 20748.06810370454 requests/second
xml to json : 17484.385980471394 requests/second
xml to json + json string : 15788.06062909986 requests/second
xml to json string : 3215.060622343105 requests/second
xml2js  : 5378.727944725807 requests/second

### DIFF
--- a/src/node2json.js
+++ b/src/node2json.js
@@ -29,6 +29,7 @@ const convertToJson = function(node, options) {
     if (node.child[tagname] && node.child[tagname].length > 1) {
       jObj[tagname] = [];
       for (var tag in node.child[tagname]) {
+        node.child[tagname].hasOwnProperty(tag) &&
         jObj[tagname].push(convertToJson(node.child[tagname][tag], options));
       }
     } else {
--- a/src/node2json.js
+++ b/src/node2json.js
@@ -28,9 +28,9 @@ const convertToJson = function(node, options) {
     var tagname = keys[index];
     if (node.child[tagname] && node.child[tagname].length > 1) {
       jObj[tagname] = [];
-      for (var tag in node.child[tagname]) {
-        node.child[tagname].hasOwnProperty(tag) &&
-        jObj[tagname].push(convertToJson(node.child[tagname][tag], options));
+      const keys = Object.keys(node.child[tagname]);
+      for (let i = 0; i < keys.length; i++) {
+        jObj[tagname].push(convertToJson(node.child[tagname][keys[i]], options));
       }
     } else {
       if(options.arrayMode === true){
--- a/spec/xmlParser_spec.js
+++ b/spec/xmlParser_spec.js
@@ -5,6 +5,9 @@ const validator = require("../src/validator");
 
 describe("XMLParser", function() {
 
+    Object.prototype.test = '';
+    Array.prototype.test = '';
+
     it("should parse all values as string, int, boolean, float, hexadecimal", function() {
         const xmlData = `<rootNode>
         <tag>value</tag>
--- a/src/json2xml.js
+++ b/src/json2xml.js
@@ -176,7 +176,8 @@ function replaceCDATAarr(str, cdata) {
     return str + '<![CDATA[' + cdata.join(']]><![CDATA[') + ']]' + this.tagEndChar;
   } else {
     for (let v in cdata) {
-      str = str.replace(this.options.cdataPositionChar, '<![CDATA[' + cdata[v] + ']]>');
+      cdata.hasOwnProperty(v) &&
+      (str = str.replace(this.options.cdataPositionChar, '<![CDATA[' + cdata[v] + ']]>'));
     }
     return str + this.newLine;
   }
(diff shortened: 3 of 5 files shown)


amitguptagwl:
Thanks for your effort. What if we just convert it in `for(let i=0; i< length; i++)`? I believe we'll not need to do an external check then. And can you please add a test as well. It is helpful to verify the changes.

Shana-AE:
> Thanks for your effort. What if we just convert it in `for(let i=0; i< length; i++)`? I believe we'll not need to do an external check then. And can you please add a test as well. It is helpful to verify the changes.

Hello, @amitguptagwl 

first of all, I'm not a experienced developer. please point out my mistake if I'm wrong. 
I think we can use `Object.keys()` and for loop to replace the check as it's can be object or array. or some even better solution?

And I think we should take for...in loop seriously and replace all of them with other solution, I found 4 place using for...in loop. I'm not sure all of them can be replaced with `Object.keys` and common for loop.

As test case. I haven't wrote one before, but I'm willing to try it.

So, I can't guarantee the time and the code quality. If you think this is a serious bug and time and code quality is a thing, please fix it.

Thank you.

Shana-AE:
Hello, @amitguptagwl I added this two line in the front of  `describe("XMLParser", function() {` in `spec\xmlParser_spec.js`, I don't know if it's appropriate.

```js
    Object.prototype.test = '';
    Array.prototype.test = '';
```

Also, I found your [nimnjs](https://github.com/nimndata/nimnjs) has the same for...in problem during the test.

amitguptagwl:
@Shana-AE 
I can't see the new changes if you did. Yes, `for` loop works faster than `for..in` and I use it most of the time if I'm not sure about the type of the variable and I want to avoid the check. I may have missed it in some places. But it'll good to fix.

Any bug can cause issues. So let me verify and fix it.I'll keep the PR open as a reminder for me.

amitguptagwl:
Just for the learning, check the above node.js CI build. Your changes in the spec file are failing 4-5 tests as It is not the right solution. However, I'll spend some time writing the correct tests in a few days.

Shana-AE:
ok, I will keep an eye on this.  Thank you for your instruction. By the way, I found these 4-5 failing test caused by [nimnjs](https://github.com/nimndata/nimnjs) for...in loop, I encountered these failing locally too, but I changed the for...in loop in nimnjs in the node_module, then the failure is gone.

amitguptagwl:
Can you please check tests? Looks like github webflow is reporting failing tests.

Shana-AE:
@amitguptagwl 

These failing caused by for...in loop in [nmnjs](https://github.com/nimndata/nimnjs), the same reason of this bug. I added prototype to Object and Array before testing starts. so it failed. Change the for...in loop in nmnjs for preventing unexpectable bug and the failing tests would be gone.

amitguptagwl:
Let me remove nimn js from FXP.

src/node2json.js
```diff
@@ -28,8 +28,9 @@ const convertToJson = function(node, options) {
     var tagname = keys[index];
     if (node.child[tagname] && node.child[tagname].length > 1) {
       jObj[tagname] = [];
-      for (var tag in node.child[tagname]) {
-        jObj[tagname].push(convertToJson(node.child[tagname][tag], options));
+      const keys = Object.keys(node.child[tagname]);
```
eacaps, line 31: Since conditional above this is checking `node.child[tagname].length > 1` which I assume is a shorthand for checking that `node.child[tagname]` is an array, would it be better to use `for(let tag of node.child[tagname])` instead? This would apply to to `src/node2json_str.js` as well.

amitguptagwl: Yes, I feel `node.child[tagname]` can be used. For `for(let tag of node.child[tagname])`, I would prefer to go with classic for loop style everywhere.

amitguptagwl:
Do the issue still exists? Since v4 beta is live now. I'm trying to merge/close all the PRs

Issue #45, opened 2018-02-08
Why carriage return removed from values?


Delagen:
### Checklist

* [x] Are you running the latest version?
* [x] Include sample input XML 
* [x] Include actual output
* [x] Include expected output
* [x] Did you try online [tool](https://naturalintelligence.github.io/fast-xml-parser/)?
* [x] Did you [star](https://github.com/NaturalIntelligence/fast-xml-parser/stargazers) the repository for further updates? ;)

I think
```
xmlData = xmlData.replace(/\r?\n/g, " ");//make it single line
```
should be removed
but
```
function buildAttributesMap(attrStr,options){
    attrStr = attrStr.replace(/\r?\n/g, " ");
```

But with this does not removed comments
```
xmlData = xmlData.replace(/<!--[\s\S]*?-->/g, "");//Remove  comments
```
resolves this

amitguptagwl:
@Delagen Have you found any test failing because of this change? (leaving the options naming aside)

If you check line number 51 of [parser.js](https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/src/parser.js), It should be removing "carriage return" already.

amitguptagwl:
Sorry I missed your comment above. Why should it be removed?
```js
xmlData` = xmlData.replace(/\r?\n/g, " ");//make it single line
```

Delagen:
@amitguptagwl Check my PR it included test.
With this line you totally remove all CRLF from entire document.
So generic test will fail

```xml
<element><a>a
b
c</a><b>
a</b></element>
```
result with trim
```
{
"element": {
  "a": "a b c",
  "b": "a"
}
}
```

but should be
```
{
"element": {
  "a": "a\nb\nc",
  "b": "a"
}
}

```
or \r\n as in source XML

amitguptagwl:
In my opinion, both are correct until we have to convert json back to the XML. But I agree to replace CRLF in only the attribute's value with single space will keep it consistent for future changes. 

However if we go with current PR, it'll break the expected (though rare) output when multiple CDATA are present in a tag. I have found the solution by changing current RE. But give me some more time if I can make it more simple.

amitguptagwl:
ah!! it was simple only. I was doing mistake in `\s` and `\\s`

amitguptagwl:
I'll publish the changes soon after fixing online demo

Delagen:
When i make PR, all tests works.

[...]

PR #289, opened 2020-09-10
fix(parse): fix parsing of multiline tag data when xml with no whitespaces between tags

Delagen:
…

# Purpose / Goal

```xml
<root><a>1
2
3</a></root>
```
previously parsed as 
```json
{"root":{"a":"1 2\n3"}}
```
instead of
```json
{"root":{"a":"1\n2\n3"}}
```

# Type
Please mention the type of PR

* [x]Bug Fix
* [ ]New Feature

### DIFF
--- a/spec/cdata_spec.js
+++ b/spec/cdata_spec.js
@@ -4,6 +4,25 @@ const parser = require("../src/parser");
 const validator = require("../src/validator");
 
 describe("XMLParser", function() {
+    it("should parse multiline tag value when tags without spaces", function() {
+        const xmlData = `<?xml version='1.0'?><root><person>lastname
+firstname
+patronymic</person></root>`;
+        let result = parser.parse(xmlData, {
+            ignoreAttributes: false
+        });
+
+        const expected = {
+            "root": {
+                "person": "lastname\nfirstname\npatronymic"
+            }
+        };
+
+        expect(result).toEqual(expected);
+
+        result = validator.validate(xmlData);
+        expect(result).toBe(true);
+    });
     it("should parse tag having CDATA", function() {
         const xmlData = `<?xml version='1.0'?>
 <any_name>
@@ -32,7 +51,6 @@ describe("XMLParser", function() {
                 }
             }
         };
-
         let result = parser.parse(xmlData, {
             ignoreAttributes: false
         });
--- a/spec/data_spec.js
+++ b/spec/data_spec.js
@@ -101,7 +101,7 @@ describe("XMLParser", function() {
                 },
                 "a:b": 2,
                 "a-b:b-a": 2,
-                "a:c": "test&amp;\r\nтест&lt;\r\ntest",
+                "a:c": "test&amp;\nтест&lt;\ntest",
                 "a:el": "<a>&lt;<a/>&lt;b&gt;2</b>]]]]>&amp;a",
                 "c:string": {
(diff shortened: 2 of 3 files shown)


Delagen:
@amitguptagwl Please comment if it is not correct as you can

amitguptagwl:
Thanks @Delagen for your PR.

I'm wondering what will we the out put for
```xml
<tag>
This is the long text I want to place here but keeping it short for this example.
</tag>
```

It'll be surrounded by '\n' unnecessarily, I believe. I feel, '\n' and white spaces are used to format the file's content. So they should be removed.

Delagen:
No, it will not be surrounded

Delagen:
The XML specification says that \n, \r, and \r\n are treated as equivalent, and are all normalized to \n by the XML parser. If you want a CR (\r) in your content that isn't going to be junked by the XML parser in this way, you have to write it as &#x0D;

amitguptagwl:
It doesn't matter what I did. I can be wrong. I was discussing what should be better. Eg if we use this library for DOM parsing then how newline character should be treated.
For `\r\n`, I think I or someone fixed an issue occurring on windows. I can't recall it exactly.

Delagen:
So we must simply remove this line https://github.com/NaturalIntelligence/fast-xml-parser/pull/289/files#diff-ad9657a089bb054342b0e4f95dc453c5R171

Delagen:
Moreover It was I https://github.com/NaturalIntelligence/fast-xml-parser/issues/45

amitguptagwl:
Previously, parser was using regx to parse the whole XML content. Hence, I was supposed to use this line;
```js
  xmlData = xmlData.replace(/(\r\n)|\n/, " ");
```
I feel If this line is removed now then it should not fail the tests. Other than a few tests related to text. However, it may break the current user experience.

Now coming to expected behavior, whether we should replace `\r`, `\n`, `\r\n` with `\n` or ` `(space). It can be controlled by a parser property. Or we can replace it with `\n` to avoid complexity of handling CDATA content and it can be further replaced by `\n` at user side.

Is my understanding correct here?

Delagen:
This rule just replace first occurence what cause error.
Ww have 2 options:
1. Leave as is, removing this line, so lf/cr will be as is.
2. Normalize to lf.

Cdata normally does not contain binary data,  so lf/cr style does not damage it's content.

But we must not replace lf to space, it broke many things, we realize XML parser, but not HTML renderer

amitguptagwl:
I think replacing `\r\n?` with `\n` would be better.

Delagen:
I think the same )

amitguptagwl:
So can you please update the PR accordingly?

Delagen:
This realize the same, what need to update?

amitguptagwl:
The change that you have done is

```js
  xmlData = xmlData.replace(/\r?\n/g, "\n");
```

But the change that we're discussing is

```js
  xmlData = xmlData.replace(/\r\n?/g, "\n");
```

Delagen:
Both regex produce the same in common cases, but yours also replace single \r without \n after which incorrect normalization.

amitguptagwl:
As you stated previously,
>  \n, \r, and \r\n are treated as equivalent, and are all normalized to \n by

* `\r?\n` should replace , `\n`, `\r\n` with `\n` (`\r` is missing to replace, and no point of replacing `\n` with `\n`)
* `\r\n?` should replace , `\r`, `\r\n` with `\n`

Delagen:
@amitguptagwl I understand your offer to optimize regex. Fixed

Delagen:
@amitguptagwl Any release schedule?

amitguptagwl:
No plan yet. But can plan for 05Oct2020. I wanted to make v4 branch live so wanted to delay all the releases till that time.

Delagen:
@amitguptagwl almost 2 months passed (

amitguptagwl:
Sorry for the delay @Delagen But due to the Corona and some unfavorable conditions, I was away from my laptop. I will publish  changes before this weekend.

amitguptagwl:
done

PR #316, opened 2021-01-27
feat(arrayMode): support RegEx and function in arrayMode option

dantio:
# Purpose / Goal
Enhance the `arrayMode` with RegEx/Callback to be able to control which Tags should be returned as an array.
This feature enables fine grained and correct XML parsing where it's known that some tags only return an array.

See [discussion](https://github.com/NaturalIntelligence/fast-xml-parser/issues/68)

# The problem
```json
{
  "items": [
      {"name": "item1"}
   ]
}
```
Will produce:
```xml
<?xml version="1.0"?>
<items>
  <name>item1</name>
</items>
```
That's fine, but if we want to parse it back to same JSON we can now use:
```js
var jsonObj = parser.convertToJson(tObj, {
  arrayMode: /items/
});
```
This will produce expected output.

Sometimes RegEx can be complicated, thefore a cb can be used:
```js
var jsonObj = parser.convertToJson(tObj, {
  arrayMode: function(tagname) { return tagname === 'items'}
});
```

Performace should only be decreased if this option is used.

# Type
Please mention the type of PR
* [ ]Bug Fix
* [x ]New Feature

# Suite: XML Parser benchmark
validation : 27487.975686280235 requests/second
xml to json : 23590.962156918744 requests/second
xml to json + json string : 20555.733716481 requests/second
xml to json string : 3414.0281077820946 requests/second
xml2js  : 6776.869020960357 requests/second

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ Validator returns the following object in case of error;
 * **cdataTagName** : If specified, parser parse CDATA as nested tag instead of adding it's value to parent tag.
 * **cdataPositionChar** : It'll help to covert JSON back to XML without losing CDATA position.
 * **parseTrueNumberOnly**: if true then values like "+123", or "0123" will not be parsed as number.
-* **arrayMode** : When `false`, a tag with single occurrence is parsed as an object but as an array in case of multiple occurences. When `true`, a tag will be parsed as an array always excluding leaf nodes. When `strict`, all the tags will be parsed as array only.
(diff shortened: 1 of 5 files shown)


dantio:
I added a second parameter `parentTagName` to `arrayMode` cb since this is also very useful e.g.:
```xml
<?xml version="1.0"?>
<result>
<itemArray>
  <item>
    <name>item1</name>
     <similarItem>
        <item><name>item2</name></item>
     </similarItem>
  </item>
</itemArray>
</result>
```
If `arrayMode=/^item$/` is applied also to item in `similarItem` node will be in array. Since we know `itemArray` should always produce an array we can use:

```js
var jsonObj = parser.convertToJson(tObj, {
  arrayMode: function(t, parentTagName) { return parentTagName === 'itemArray'}
});

```

amitguptagwl:
it sounds interesting. And i believe it'll be helpful too. I'll keep the PR open for few more days to welcome any new thought and changes.

dantio:
Thanks! That feature is already in use in the next version of https://github.com/hendt/ebay-api :)

PR #322, opened 2021-02-26
Fix Import of nimndata in parser

aaronmgdr:
# Purpose / Goal
On a fresh install off this package whenever I tried to use the parse I would run into a an error 

```
  Could not locate module ../src/nimndata mapped as:
    /Users/{my project...}/src/$1.
```

I notice that all the other imports in the `parse.js` file use `"./filename"` for imports. So i changed this to match. 

# Type
* [x] Bug Fix


aaronmgdr:
if this is intentional happy to leave

PR #371, opened 2021-10-01
Better error location for unclosed tags

### LINKED ISSUE #370: Unclosed tag in xml: Line number reported by validate() should be the opening tag's line

### Description
When there's an unclosed xml tag in the input, the `validate()` method does not output the line of the unclosed tag. Instead, the line of the parent's closing tag is reported as the error location. This is understandable, because this is the location where it's certain that there is an error (other parsers do the same). But it's not very helpful in finding the input error.

If possible, the error message should be changed to point to the unclosed tag.

### Code
```js
const fastXmlParser = require('fast-xml-parser');

const xml =
`<parent>
  <child name="Peter"/>
  <child name="Paul">      <!-- ERROR: should be self-closing -->
  <child name="Mary"/>
</parent>`;

console.log(fastXmlParser.validate(xml));
```

### Output
```js
{
  err: {
    code: 'InvalidTag',
    msg: "Closing tag 'child' is expected inplace of 'parent'.",
    line: 5   // <<<< not helpful
  }
}
```
### Desired Output
```js
{
  err: {
    code: 'InvalidTag',
    msg: "Unclosed tag 'child'",
    line: 3       // <<<< line of the opening tag
  }
}
```

- [x] Yes
- [ ] No

Gei0r:
Fixes #370

# Purpose / Goal
For unclosed tags, `validate` now reports the opening tag's line instead of the line where the parent is closed. See #370.

I only changed `src/validator.js`. I tried re-building `lib/`, but `yarn run bundle` creates a large number of changed files, so I'm not sure if that is the correct call.

# Type
Please mention the type of PR

* [x]Bug Fix
* [ ]New Feature

### DIFF
--- a/src/validator.js
+++ b/src/validator.js
@@ -35,7 +35,7 @@ exports.validate = function (xmlData, options) {
     }else if (xmlData[i] === '<') {
       //starting of tag
       //read until you reach to '>' avoiding any '>' in attribute value
-
+      let tagStartPos = i;
       i++;
       
       if (xmlData[i] === '!') {
@@ -105,8 +105,8 @@ exports.validate = function (xmlData, options) {
             return getErrorObject('InvalidTag', "Closing tag '"+tagName+"' can't have attributes or invalid starting.", getLineNumberForPosition(xmlData, i));
           } else {
             const otg = tags.pop();
-            if (tagName !== otg) {
-              return getErrorObject('InvalidTag', "Closing tag '"+otg+"' is expected inplace of '"+tagName+"'.", getLineNumberForPosition(xmlData, i));
+            if (tagName !== otg.tagName) {
+              return getErrorObject('InvalidTag', "Unclosed tag '"+otg.tagName+"'.", getLineNumberForPosition(xmlData, otg.tagStartPos));
             }
 
             //when there are no more tags, we reached the root level.
@@ -127,7 +127,7 @@ exports.validate = function (xmlData, options) {
           if (reachedRoot === true) {
             return getErrorObject('InvalidXml', 'Multiple possible root nodes found.', getLineNumberForPosition(xmlData, i));
           } else {
-            tags.push(tagName);
+            tags.push({tagName, tagStartPos});
           }
           tagFound = true;
         }


amitguptagwl:
Thanks for your changes. Can you please add a test for the same?

Gei0r:
Done, please take a look.

amitguptagwl:
Let's take a few more examples and understand if we need any change in the fix

```xml
<parent>
  <child>
     <self/>
  </incorrect>
  <empty>
<parent>
```

Gei0r:
Your example produces:

```json
{
  "err": {
    "code": "InvalidTag",
    "msg": "Unclosed tag 'child'.",
    "line": 2
  }
}
```

Now the question is whether this is as expected. 

I could change the validation output, so that it looks through the list of opened tags. 

- If there is a tag closure but no corresponding tag opening (as in your example), the error message could be changed to something like `expected closing tag 'child' (opened in line 2) instead of closing tag 'incorrect'` and `line = 4`.
- If there is a tag closure but somewhere in the tag list there is a corresponding tag opening (as in my example), the error message would just say `Unclosed tag 'childB'.`

This way, the code would differentiate between "forgot to close tag" and "closed tag incorrectly".

Let me know what you think.

amitguptagwl:
When a tag is never closed then specifying the line number of starting tag makes sense. In other cases, specifying the line number of both opening and closing tag would be good where the line number separately is pointing the line where the issue is.

But for the other case you're talking aboout, how would you differentiate? Eg
```xml
<parent>
  <child>
     <self/>
     <self>
  </incorrect>
  <empty>
<parent>
```

Gei0r:
`validate()` will report the first error it finds. So in your example it would produce `Unclosed tag 'self'` in `line=4`.

If you fix line 4, the next error will be `expected closing tag 'child' (opened in line 2) instead of closing tag 'incorrect'` in `line=5`.

amitguptagwl:
Can you please remove browser bundle files from the PR? I prefer to generate them at the time of release.

Issue #403, opened 2021-12-05
XML Builder formatted XML


amitguptagwl:
Output of XML Builder when `preserveOrder:true` needs to be improved.

UltraCakeBakery:
I was about to create an issue about preserveOrder not preserving order correctly after build. It also makes the wrong elements children of the wrong parents. Should I make an issue about this or do you think you know what I'm talking about?

amitguptagwl:
@UltraCakeBakery if it is just related to format then you may skip it. And it is on less priority. But if it is related to functionality, please raise another issue with examples. I'll look into.

urav06:
I am working on this but I cant seem to work the `orderedJs2Xml.js`
when I run the `XMLBuilder` with `preserveOrder` set to `true` on the test input `/spec/assets/test_ordered.json` , it returns an empty string. Is this expected?

dantio:
I'm not able to use `preserveOrder:true` with `XMLBuilder`. This option apparently required to build xml with CDATA.
What I'm trying to do:
```js
 const options = {
      preserveOrder: true,
      cdataPropName: '__cdata',
    }
const builder = new XMLBuilder(options);
const result = builder.build({
        "Item": {
            "__cdata": "<p></p>"
        }
     });
```
The result is:
`<?xml version="1.0"?>`

With `preserveOrder:false` the result is:
```xml
<?xml version="1.0"?>
<Item>
  <__cdata>&lt;p&gt;&lt;/p&gt;</__cdata>
</Item>
```

Actually I think XMLBuilder does not work with  `preserveOrder:true`  at all...

amitguptagwl:
This thread was created only for formatting of XML. So closing this.

Issue #421, opened 2022-01-13
Add Configuration List for SuppressEmptyTags


diomedtmc:
* [x] Are you running the latest version?
* [x] Have you included sample input, output, error, and expected output? 
* [x] Have you checked if you are using correct configuration?
* [x] Did you try online [tool](https://naturalintelligence.github.io/fast-xml-parser/)?

### Description
We often need to parse/build HTML and not traditional XML with this tool.  However, in HTML world, blunt force on the empty tags is not good.  We would like to see an empty div, span, p, etc...remain in tact with a close and open tag.  However, for say a br or hr, we would rather suppress it.  Would love to see a configurable list for which tags we suppress in the XmlBuilder instead of a global boolean.

### Input
1.  Create a simple XML string with an empty tag you DO NOT want to suppress and one you do want to supress.
2.  Use XmlParser
3.  Use an XmlBuilder to regenerate the XML from the output in step 2

### Code
```js
const sample = '<article><div style="padding: 20px"></div><br/></article>';
const sharedFastXmlConfiguration = {};

const parser = new XmlParser(sharedFastXmlConfiguration);
const builder = new XmlBuilder(sharedFastXmlConfiguration);
const elements = parser.parse(sample);
const sampleAfterFastXmlParser = builder.build(elements);

expect(sampleAfterFastXmlParser).to.equal(sample);
```

### Output
Depending on how you configure it, you end up with either (suppress empty tags true):

```js
const sampleAfterFastXmlParser = '<article><div style="padding: 20px"/><br/></article>';
```

Or(suppress empty tags true):

```js
const sampleAfterFastXmlParser = '<article><div style="padding: 20px"></div><br><br/></article>';
```

### expected data
Instead, I should be able to get closer to proper HTML (and conceivably match my input):

```js
const sampleAfterFastXmlParser = '<article><div style="padding: 20px"></div><br/></article>';
```

**Would you like to work on this issue?**

- [ ] Yes
- [x] No

[Bookmark](https://github.com/NaturalIntelligence/fast-xml-parser/stargazers) this repository for further updates.

amitguptagwl:
Have you tried `unpairedTags: ["hr", "br", "link", "meta"]` with `suppressEmptyNode: false`?

Ref: [HTML Parsing with FXP](https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/docs/v4/6.HTMLParsing.md)

diomedtmc:
[...]

PR #439, opened 2022-03-07
Implement API to pretty format XML strings

bajtos:
# Purpose / Goal

I am looking for an easy to use API for formatting (pretty-printing) XML. Something like `xmllint --format`.

fast-xml-parser already has both XMLParser and XMLBuilder classes that can preserve all bits and pieces (comments, processing instructions, etc.), but it's not clear what settings to use for pretty printing.

In my pull request, I am adding a simple function accepting an XML string and returning back the formatted version. Under the hood, the function configures XMLParser and XMLBuilder in such a way so that as much XML data is preserved as possible.

### Example use:

```js
const {format} = require('fast-xml-parser');
const result = format('<data><!-- comment --><value>1</value></data>');

/* result:

<data>
  <!-- comment -->
  <value>
    1
  </value>
</data>

*/
```

Related:
- #403
- https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/docs/v4/3.XMLBuilder.md#restoring-original-xml

# Type
Please mention the type of PR

* [ ] Bug Fix
* [x] New Feature

### DIFF
--- a/spec/formatter_spec.js
+++ b/spec/formatter_spec.js
@@ -1,0 +1,58 @@
+'use strict';
+
+const {format} = require('../src/fxp');
+
+describe('XML formatter', () => {
+  it('should format simple document', () => {
+    const result = format('<data><item>1</item></data>');
+    expect(result).toBe(`
+<data>
+  <item>
+    1
+  </item>
+</data>`);
+  });
+
+  it('should preserve XML PI tag', () => {
+    const result = format('<?xml version="1.0"?><data>hello</data>');
+    expect(result).toBe(`
+<?xml   version="1.0"?>
+<data>
+  hello
+</data>`);
+  });
+
+  it('should preserve self-closed tags', () => {
+    const result = format('<data><self-closed/></data>');
+    expect(result).toBe(`
+<data>
+  <self-closed/>
+</data>`);
+  });
+
+  it('should preserve comments', () => {
+    const result = format('<data><!-- comment --><value>1</value></data>');
+    expect(result).toBe(`
+<data>
+  <!-- comment -->
+  <value>
+    1
+  </value>
+</data>`);
+  });
+
+  it('should preserve attributes', () => {
+    const result = format('<data kind="text">hello</data>');
+    expect(result).toBe(`
+<data kind="text">
+  hello
+</data>`);
+  });
+
+  xit('should preserve closing tag after empty content', () => {
+    const result = format('<empty></empty>');
+    expect(result).toBe(`
+<empty>
+</empty>`);
+  });
+});
--- a/src/formatter.js
+++ b/src/formatter.js
@@ -1,0 +1,26 @@
+'use strict';
+
+const XMLParser = require('./xmlparser/XMLParser');
+const XMLBuilder = require('./xmlbuilder/json2xml');
+
+exports.format = function(xmlString) {
+  const options = {
+    alwaysCreateTextNode: true,
+    cdataPropName: '#cdata',
+    commentPropName: '#comment',
+    ignoreAttributes: false,
+    preserveOrder: true,
+    processEntities: false,
+    trimValues: false,
+    suppressEmptyNode: true,
+    format: true,
+  };
+
+  const parser = new XMLParser(options);
+  const data = parser.parse(xmlString);
+
+  const builder = new XMLBuilder(options);
+  const result = builder.build(data);
+
+  return result;
+};
--- a/src/fxp.d.ts
+++ b/src/fxp.d.ts
@@ -86,3 +86,5 @@ export class XMLBuilder {
   constructor(options: XmlBuilderOptionsOptional);
   build(jObj: any): any;
 }
+
+export function format(xmlString: string): string;
--- a/src/fxp.js
+++ b/src/fxp.js
@@ -3,9 +3,11 @@
 const validator = require('./validator');
 const XMLParser = require('./xmlparser/XMLParser');
 const XMLBuilder = require('./xmlbuilder/json2xml');
+const {format} = require('./formatter');
 
 module.exports = {
   XMLParser: XMLParser,
   XMLValidator: validator,
-  XMLBuilder: XMLBuilder
-}
+  XMLBuilder: XMLBuilder,
+  format,
+};


src/formatter.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const XMLParser = require('./xmlparser/XMLParser');
+const XMLBuilder = require('./xmlbuilder/json2xml');
+
+exports.format = function(xmlString) {
```
bajtos, line 6: Discussion points:

1. What is a good name for this function? Do we want to use `XML` prefix for consistency with other exports (e.g. `XMLParser`)?
2. Is it ok to use a function instead of a class? Personally, I prefer to use the simplest solution that works. However, I see how a class-based API can be better for consistency.

    ```js
    const formatter = new XMLFormatter();
    const result = formatter.format('<data>hello</data>');
    ```

3. Do we want to add an `options` argument, to allow callers to tweak certain aspects of parsing & building? Example candidates: `trimValues` and `suppressEmptyNode`?

src/fxp.d.ts
```diff
@@ -86,3 +86,5 @@ export class XMLBuilder {
   constructor(options: XmlBuilderOptionsOptional);
   build(jObj: any): any;
 }
+
+export function format(xmlString: string): string;
```
bajtos, line 90: I haven't written any documentation yet, I am waiting until we agree on the desired API.

What is your recommendation? Should I create another file in `docs` directory, e.g. `docs/formatter.md`? Maybe we can add JSDoc comments to this `fxp.d.ts` file too, to allow IDEs like VS Code to show inline docs?

spec/formatter_spec.js
```diff
@@ -0,0 +1,58 @@
...
+const {format} = require('../src/fxp');
+
+describe('XML formatter', () => {
+  it('should format simple document', () => {
+    const result = format('<data><item>1</item></data>');
+    expect(result).toBe(`
+<data>
+  <item>
+    1
+  </item>
```
bajtos, line 12: I find this choice a bit surprising. Personally, I'd prefer the text to be on the same line as the opening tag.

Here is an example showing how `xmllint` handles formatting:

```
$ echo "<data><item>1</item></data>" | xmllint --format -
<?xml version="1.0"?>
<data>
  <item>1</item>
</data>
```

Anyhow, I think changing the formatting rules is out of scope of this pull request. If you are ok with changing the formatter to make it closer to `xmllint`, then I am happy to open another pull request.

spec/formatter_spec.js
```diff
@@ -0,0 +1,58 @@
...
+<data kind="text">
+  hello
+</data>`);
+  });
+
+  xit('should preserve closing tag after empty content', () => {
+    const result = format('<empty></empty>');
+    expect(result).toBe(`
+<empty>
+</empty>`);
```
bajtos, line 56: I could not find a way how to make this test pass, while keeping `<self-closed/>` test to pass too. When `suppressEmptyNode` is enabled, `<self-closed/>` passes `<empty></empty>` fails. When I disable that setting, then the first test fails and this one passes.

If we want to support both cases, then I think we need to improve the parser to store metadata allowing the builder to distinguish between a self-closed element and an element with empty content.

WDYT?

spec/formatter_spec.js
bajtos: I see this has been discussed in https://github.com/NaturalIntelligence/fast-xml-parser/discussions/401 with the following conclusion:

> It's difficult to decide when opening and closing tags should be in-lined text data. However, we can provide a configuration for this.

Can you @amitguptagwl please show me some pointers on how & where to start implementing such configuration?

amitguptagwl:
First of all thanks to raise this PR. Let me try to understand your purpose.

Currently, the XML formatter is embedded with XML builder. It is not external. Hence, not much impact on performance. Are you trying to externalize it?

bajtos:
Thank you @amitguptagwl for a quick reply 👍🏻 

> Currently, the XML formatter is embedded with XML builder. It is not external. Hence, not much impact on performance. Are you trying to externalize it?

I am looking for an easy to use API for formatting XML. I can manually call `XMLParser` and `XMLBuilder` myself (duplicating the code in [`src/formatter.js`](https://github.com/bajtos/fast-xml-parser/blob/c2b0a11d4c58c41aeb50fda7cd4b0ab9a619db5f/src/formatter.js)), I find such use rather cumbersome.

I did not realize there is already a XML formatter inside the XML builder. I am NOT trying to externalize it. Based on your comment, I think my choice to use "formatter" name was not ideal. Maybe "pretty printer" would be a better name?

The purpose of my PR is to give users of fast-xml-parser an easy way how to (re)format a XML string for easier reading by humans, as shown in the Example use in the PR description.

amitguptagwl:
Since, you're not planning to externalize it then probably using `format: true` property with builder solve the purpose (as today) ?

Here, is the documentation of [XML builder](https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/docs/v4/3.XMLBuilder.md#format).

Please let me know if I'm missing something to understand this change.

bajtos:
> Since, you're not planning to externalize it then probably using `format: true` property with builder solve the purpose (as today) ?

Yes and no.

Yes, adding `format: true` makes XMLBuilder print the XML in a pretty format.

It's not enough though, we also need to configure XMLParser to preserve data needed to reconstruct the original XML document, e.g. `ignoreAttributes: false` and `preserveOrder: true`.

> Please let me know if I'm missing something to understand this change.

Yeah, it feels like I am not able to get you to understand my intention here. Let me try again.

At the moment, fast-xml-parser provides all building blocks necessary to write a function that will accept a XML string and return back a nicely formatted version of that document.

However, to write such a function, one has to understand many low-level details about the way how XMLParser and XMLBuilder works, and configure them correctly.

My intention is to improve fast-xml-parser by implementing a new API, a wrapper function, that hides all of this complexity under the hood and provides an easy to use API.

Now I have to write something like this, and I am not sure if all of these options are necessary and whether I did not miss anything:

```js
const options = {
  alwaysCreateTextNode: true,
  cdataPropName: '#cdata',
  commentPropName: '#comment',
  ignoreAttributes: false,
  preserveOrder: true,
  processEntities: false,
  trimValues: false,
  suppressEmptyNode: true,
  format: true,
};

const parser = new XMLParser(options);
const data = parser.parse(xmlString);

const builder = new XMLBuilder(options);
const result = builder.build(data);
```

If you accept my pull request, then I could simplify my code as follows:

```js
import {format} from 'fast-xml-parser';
const result = format(xmlString);
```

It's all about developer experience and ease of use.

As a nice side effect, by having tests for the parse-build roundtrip, we can more easily find out how various edge cases are handled, and perhaps improve the implementation later. For example, in the tests in this PR, I found it's not possible to preserve both self-closed and empty tags right now: `<data><self-closed/><empty></empty></data>`.

bajtos:
I am arguing that fast-xml-parser users should not be required to understand the complexity of parser & builder options. It's something that should be handled by fast-xml-parser authors, because they have the best understanding of how different options are affecting the parse-build roundtrip.

amitguptagwl:
Probably we both have different perspective in our mind so not able to reach to a point. But let me clarify few things here.

A user don't need to remember any configuration just for formatting. But she need to remember the configure if they that particular feature from parser or builder. So if the user doesn't want particular feature from a parser, they don't need to set the same in builder. Eg "comment" property is required to be set in parser if user don't want to skip the comments from the result. Similarly, if he is including the comments in parser result then he may want to set the same property to builder so that the builder doesn't build the comment as XML tag. However, it is completely optional and not related to formatting.

Formatting needs very minimum options and they're set to their default values. Eg
* format
* indentBy

There are 3 more options which might be confused with formatting but they're basically deciding the output.
* suppressUnpairedNode
* suppressBooleanAttributes
* suppressEmptyNode

So I feel what we may need is good documentation with good examples. Please let me know if I'm able to express my view correctly. Or I'm fine to discuss it more.

amitguptagwl:
One more point, if I'm understanding your point correctly with this code
```js
import {format} from 'fast-xml-parser';
const result = format(xmlString);
```

The purpose of FXP is not to parse & build a XML string back to the XML string.

bajtos:
> A user don't need to remember any configuration just for formatting. But she needs to remember the configure if she wants that particular feature from parser or builder. So if the user doesn't want particular feature from a parser, they don't need to set the same in builder. Eg "comment" property is required to be set in parser if user doesn't want to skip the comments from the result. Similarly, if she is including the comments in parser result then she may want to set the same property to builder so that the builder doesn't build the comment as XML tag. However, it is completely optional and not related to formatting.

I see your point.

As a user that wants to reformat an XML document, preserving all bits and pieces, I find it too complex that I have to tell the parser to preserve comments, emit attributes, and so on. It's easy to miss an option like `preserveOrder` that may not be important in sample data but could backfire in production when different XML comes in.

That's why I am proposing to add a new API for reformatting documents so that we can hide that complexity from the user and give them a well-tested API that preserves all parts of the input document.

> I feel what we may need is good documentation with good examples

Having documentation explaining how to pretty format XML documents would be definitely an improvement. The documentation has a downside though - it requires extra effort to write automated tests to ensure the documented configuration works as intended, handles correctly all edge cases, and future changes are not going to break that.

> The purpose of FXP is not to parse & build a XML string back to the XML string.

I think this is the most important point! I totally understand your desire to keep FXP features & the API surface small. I know from my own experience that a smaller feature set makes packages easier to maintain.

I can implement my `format` function as a new package depending on FXP. But:

The problem with such an approach is that we (the npm ecosystem) will have yet another XML-related package, yet another dependency to add to our projects, with the risk that it may become unmaintained in the future. I did a detailed search of XML-related packages before I choose FXP, and there are already too many of them. I don't want to make the problem even worse!

I was hoping that if we put the `format` function into FXP, it will make it easier for developers to find a decent implementation of "pretty format XML" because it will come as a built-in part of the popular FXP package.

Another point to consider: many formatters depend on an XML parser, usually a different one than FXP. I am using FXP in other places of my codebase for parsing, it works great for me! Now if I pick a 3rd-party formatter, then I'll end up with two XML parsers in my dependency tree. That seems silly to me.

Having said that, I understand your priorities may be different as a maintainer of this package.

Do you have any recommendations to package consumers on which package to use for pretty formatting XML documents, if FXP does not want to support that use case? In other words, when people have a project that's using FXP for parsing, what solution do you recommend for pretty-printing XML?

amitguptagwl:
Let's talk in terms of real uses and examples,

1. User A wants the FXP to parse comments to take some action but they should not appear in builder.
2. User B wants the FXP to parse comments and PI instructions and he want to restore the same XML with some changes and without any change.
3. User C wants to convert unpaired tags from HTML to self-closing tag and remove comments.

There can 1000s of combinations. These are just 3 examples. Let me know please how will you cover them.

> I find it too complex that I have to tell the parser to preserve comments, emit attributes, and so on.

No, there are already some default settings. But since 1000s of the combinations are possible, users can't ignore them completely.

> It's easy to miss an option like `preserveOrder` that may not be important in sample data but could backfire in production when different XML comes in.

People should not even remember that. It was made only for the users who want to preserve the sequence. It completely change the output structure and looks ugly. This is only suitable for the people who want to many modifications in the parsed result.

> when people have a project that's using FXP for parsing, what solution do you recommend for pretty-printing XML?

I tried to explore that to suggest you but couldn't find something good. So you'll have to probably find something good that we can suggest to more users as well.

----

If your requirement is limited to just parse a unformatted XML string without any other change like `<data><item>1</item></data>` to 
```xml
<data>
    <item>1</item>
</data>
```
then FXP is the bad choice. It'll reduce the formatting performance by 3 times I believe, depends on the complexity of the XML document.

And also check [this](https://github.com/NaturalIntelligence/fast-xml-parser/issues/421#issuecomment-1013607141) example which is related to only formatting. I would be happy to know your solution on that.

----

Let's come to some options that you opted to set as default

* `alwaysCreateTextNode: true`: This is not required for the examples you suggested.
* `cdataPropName`: This property is currently not supported until some new changes go on.
* `ignoreAttributes`: A default option could have been set to `true` but it was knowingly kept `false` as most of the developers don't use attributes for small XMLs while parsing from some configurations, and tests. So you can say it is based on need.
* `preserveOrder` is created only if sequence matters. So it can be set to `true` by default but it'll break the user experience of old users migrating to v4.
* `processEntities: false`: it must be set to `true` as supported by all XML parsers, browsers, and need.

amitguptagwl:
@bajtos I'm closing this PR. You can probably create a Discussion topic to discuss it further.

Issue #479, opened 2022-07-24
Playground does not work


kylemh:
<img width="1728" alt="Screen Shot 2022-07-25 at 12 20 33 AM" src="https://user-images.githubusercontent.com/9523719/180666301-573d14aa-1089-4d12-836e-f64d28dedc4b.png">

The URL listed here (https://naturalintelligence.github.io/fast-xml-parser/) does not validate or parse XML, nor convert JSON/JS into XML.

amitguptagwl:
Please reopen if still exists

PR #486, opened 2022-08-16
Update CodeMirror CDN Links in index.html

layaxx:
TLDR: Demo site was broken due to missing dependency

# Purpose / Goal
Demo site for this package is broken, see e.g. https://github.com/NaturalIntelligence/fast-xml-parser/issues/479

This is due to outdated CDN links for the dependency CodeMirror (giving 404 responses). I replaced the broken links for both javascript and CSS files with links to a working CDN.

# Type
Please mention the type of PR

* [x] Bug Fix
* [ ] New Feature

### DIFF
--- a/index.html
+++ b/index.html
@@ -18,10 +18,16 @@
         <link rel="stylesheet" href="static/css/bootstrap.min.css">
         <link rel="stylesheet" href="static/css/prettify.min.css">
         <link rel="stylesheet" href="static/css/jquery-confirm.min.css">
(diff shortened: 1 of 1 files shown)


layaxx:
Also worth noting that the demo generates invalid XML (one additional `<?xml></?xml>` tag) for the default input if one simply parses is to json and back to xml with the default config. Not sure if that is an issue with the demo or with the library itself.

PR #499, opened 2022-09-25
Add support for ignoring < inside stop nodes

nullcatalyst:
This is useful for parsing
```
<script>if (a < b) {}</script>
```
since without it, the < is treated as a new tag. This causes the parser to throw an exception when it cannot find a corresponding > to close the tag

# Purpose / Goal
Ideally I'd like to use this library to parse a small subset of HTML, and one of the problems I found with it is that it unfortunately doesn't handle `<` in script tags particularly well.

Currently it treats `<` as the start of an open tag, then looks for a corresponding `>` to close that tag. Sometimes this is a desirable feature to have -- like when parsing `<pre>` tags, the contents should still be valid HTML.

It's just `<script>` that is the odd ball here. This is also why it cannot simply be applied to all `options.stopNodes`, so a new options field had to be added. I went with `options.ignoreTagsInNodes` though I am not attached to that name at all, if you have a better one, please change it :)

# Type
Please mention the type of PR
* [ ]Bug Fix
* [x]New Feature

### DIFF
--- a/spec/html_spec.js
+++ b/spec/html_spec.js
@@ -78,4 +78,66 @@ const parsingOptions = {
     output = output.replace('₹','&inr;');
   expect(output.replace(/\s+/g, "")).toEqual(html.replace(/\s+/g, ""));
     });
+
+
+    it("should fail to parse HTML <script> tags containing '<' without using options.ignoreTagsInNodes", function () {
+        const html = `
+            <html lang="en">
+                <body>
+                    <script>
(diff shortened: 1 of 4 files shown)


amitguptagwl:
Let me understand your changes. I'll try to respond as soon as possible from my side.

amitguptagwl:
Probably, we can rename `checkNodePathMatch` to `isNodeInPaths`.

amitguptagwl:
@nullcatalyst I believe this is covered in [@nodable/flexible-xml-parser](https://www.npmjs.com/package/@nodable/flexible-xml-parser)

Issue #490, opened 2022-08-19
IsArray option isn't parsing tags with 0 as value correctly


DolarJoe:
* [ ] Are you running the latest version?
`"fast-xml-parser": "^4.0.9"`
* [x] Have you included sample input, output, error, and expected output? 
* [x] Have you checked if you are using correct configuration?
* [x] Did you try online [tool](https://naturalintelligence.github.io/fast-xml-parser/)?
It seems to be broken

### Description
I'd like to use IsArray option, however, it seems to break on tags which have value 0 inside of them, such as `0` or `0.0`. It is inconsistent with other numbers, interger or floats, that have non 0 value, as shown in my provided example.
### Input

```
<root>
    <a>0.00</a>
    <b>0.01</b>
</root>
```

### Code
```
const parser = new XMLParser({isArray: (name, jpath, isLeaf, isAttr) => !isLeaf});
let result = parser.parse(fr.result);
console.log(JSON.stringify(result, undefined, 2))
```

### Output
```
{
  "root": [
    {
      "a": [
        0
      ],
      "b": 0.01
    }
  ]
}
```

### expected data

```
{
  "root": [
    {
      "a": 0,
      "b": 0.01
    }
  ]
}
```
**Would you like to work on this issue?**

- [ ] Yes
- [x] No

omggga:
Yes, the same issue for me.

There is a function for it in code:

```
function isLeafTag(obj, options){
  const propCount = Object.keys(obj).length;
  if( propCount === 0 || (propCount === 1 && obj[options.textNodeName]) return true;
  return false;
}
```

https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/src/xmlparser/node2json.js#L98

The problem is that `obj[options.textNodeName]` is checking the value exist, but if it `false` or `0` it will fail the test.
So you can try to fix it to test it also for boolean type or zero.

You can check it with this xml:
```
<root>
    <a>00</a>
    <b>false</b>
</root>
```

and this option:

```
const isArray = function(name, jpath, isLeafNode, isAttribute){ 
	if (!isLeafNode) return true
}
```

amitguptagwl:
fixed as part of 4.1.4

PR #557, opened 2023-03-21
IsArray option isn't parsing tags with 0 as value correctly #490

omggga:
# Purpose / Goal
This is fix for isArray option isn't parsing tags with 0 as value correctly. This only affected when we are parsing tags, not attributes.

The problem is that `obj[options.textNodeName]` is checking the value exist, but if it `false` or `0` it will fail the test.
So you can try to fix it to test it also for boolean type or zero.

The opened issue is #490 

# Type
Please mention the type of PR

* [x]Bug Fix
* [ ]New Feature

### DIFF
--- a/spec/arrayMode_spec.js
+++ b/spec/arrayMode_spec.js
@@ -175,6 +175,37 @@ describe("XMLParser with arrayMode enabled", function () {
         expect(result).toEqual(expected);
     });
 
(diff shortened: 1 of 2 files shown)


amitguptagwl:
Can you please add a test for the same?

omggga:
> Can you please add a test for the same?

I added a test to check this feature.

PR #574, opened 2023-05-13
Make options in objects optional

Lenni009:
# Purpose / Goal
When using only two options in an options object which has 20 possible options, it's a bit too much to add all of them when you only need two. So now most of them have been made optional.
TypeScript would complain if you didn't use all 20, but used the corresponding type for that object.

# Type
Please mention the type of PR
* [x]Bug Fix
* [ ]New Feature


Lenni009:
Might actually be not necessary, since TS can infer the types for the options on its own

PR #582, opened 2023-05-31
fix: return type for `tagValueProcessor` & `attributeValueProcessor`

### LINKED ISSUE #581: Incorrect return type for `tagValueProcessor` and `attributeValueProcessor` in XmlBuilderOptions

### Description
Both `tagValueProcessor` and `attributeValueProcessor` are declared as returning a string, but the function should allow returning the value as is (which is also [the default](https://github.com/NaturalIntelligence/fast-xml-parser/blob/c7b3cea4ead020c21d39e135a50348208829e971/src/xmlbuilder/json2xml.js#L16-L21)).

### Code
```ts
const xmlBuilder = new XMLBuilder({tagValueProcessor: (name, value) => value});
```

### Output
```
Type '(name: string, value: unknown) => unknown' is not assignable to type '(name: string, value: unknown) => string'.
  Type 'unknown' is not assignable to type 'string'.ts(2322)
fxp.d.ts(65, 3): The expected type comes from property 'tagValueProcessor' which is declared here on type 'Partial<XmlBuilderOptions>'
```

### expected data
No typescript error.

- [x] Yes
- [ ] No

monholm:
Resolves #581

# Purpose / Goal
Addresses #581

# Type
Please mention the type of PR

* [x]Bug Fix
* [ ]New Feature


monholm:
Fixed the merge conflicts. With the recent addition of JSDocs, the discrepancy has become event more apparent as the JSDocs and TS types doesn't match.

@amitguptagwl anything I can do to help move this along? :)

amitguptagwl:
Thanks for the correction

PR #594, opened 2023-06-23
fix(types): make `build` and `parse` generic

sarahdayan:
# Purpose / Goal

This PR improves typing for `build` and `parse` by making both functions generic.

It also fixes the return type of `build`, which should be `string`.

## `parse`

```ts
// Before
const result = xml.parse(data); // `result` is `any`

// After
type ExpectedObject = {
  title: string;
  description: string;
};

const result = xml.parse<ExpectedObject>(data); // `result` is `ExpectedObject`
```

## `build`

```ts
// Before
const xml = builder.build(obj); // `xml` is `any`

// After
const xml = builder.build(obj); // `xml` is `string`
```

# Type
Please mention the type of PR

* [x] Bug Fix
* [ ] New Feature

### DIFF
--- a/src/fxp.d.ts
+++ b/src/fxp.d.ts
@@ -88,9 +88,11 @@ type ValidationError = {
   };
 };
 
+type GenericObject = Record<string, unknown>;
+
 export class XMLParser {
   constructor(options?: X2jOptionsOptional);
-  parse(xmlData: string | Buffer ,validationOptions?: validationOptionsOptional | boolean): any;
+  parse<TValue extends GenericObject = GenericObject>(xmlData: string | Buffer ,validationOptions?: validationOptionsOptional | boolean): TValue;
   /**
    * Add Entity which is not by default supported by this library
    * @param entityIndentifier {string} Eg: 'ent' for &ent;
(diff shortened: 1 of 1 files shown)


amitguptagwl:
Can you please explain me your changes for parse function? What exactly we're expecting

sarahdayan:
@amitguptagwl This is only a type change. `parse` used to return `any`, forcing you to cast it. Now you can provide a generic type to maintain type-safety, which is usually preferrable as many linters frown upon type casting.

amitguptagwl:
What I can understand here is to define the return type of parser to `ExpectedObject` with properties `title` and `description` which is not true for parser. Hence, I'm confusd what exactly you mean with the type you defined.

equipatuequipo:
In the example they gave the type `ExpectedObject` has the properties `title` and `description` but in a real world use that type have the properties that you expect the parsed XML to have. 
For example for this XML:
```xml
<person>
     <name>John</name>
     <surname>Doe</surname>
</person>
```
the corresponding Typescript type would be:
```typescript
type Person = {
    name: string;
    surname: string;
}
```
and you would use it like:
```typescript
const result = xml.parse<Person>(data);
```
This would make the `result` variable have the type `Person` instead of `any` which is much better in terms of type safety.
You can read more about generics [here](https://www.typescriptlang.org/docs/handbook/2/generics.html).

PR #595, opened 2023-06-26
Remove trailing slash from jPath for self-closing tags

### LINKED ISSUE #564: jPath for self-closing XML tag includes slash
* [x] Are you running the latest version? - yes, 4.2.2
* [ ] Did you try online [tool](https://naturalintelligence.github.io/fast-xml-parser/)?

### Description
 

jPath in `updateTag()` callback for self closing tag ends with `/`, which I doubt is intended.

Because of this, you need to check two different jPaths, with and without `/`, to accurately process tags that may be empty.

### Input

```xml
<doc>
	<foo/>
	<foo>bar</foo>
	<lorem>ipsum</lorem>
</doc>
```

### Code

```js
const skip = ["doc.foo"];
	updateTag: (tagName, jPath) => {
		if (skip.includes(jPath)) {
			return false;
		}
		return tagName;
	},
});
```

### Output

```json
{
  "doc": {
    "foo": "",
    "lorem": "ipsum"
  }
}
```

### expected data

```json
{
  "doc": {
    "lorem": "ipsum"
  }
}
```

### additional info

Code works if the `skip` array is changed to: `["doc.foo", "doc.foo/"]` because for the self-closing tag jPath ends with `/`.
But I think that's easy to miss and not intended to be so? If you write code wanting to skip particular path and you don't test it with both self-closing and not-self closing tag you will be surprised.

Another workaround is to do:

```js
	updateTag: (tagName, jPath) => {
		if (
			skip.includes(jPath) ||
			(jPath.endsWith("/") && skip.includes(jPath.slice(0, -1)))
		) {
			return false;
		}
		return tagName;
	},
```

- [x] Yes
- [ ] No

m-radzikowski:
# Purpose / Goal

Fixes #564 

The `jPath` in `updateTag()` callback for self-closing tags contained a trailing slash (like: `root.a/`) which was not intended and could lead to errors if you implemented logic based on the `jPath` and did not take that into account.

I've added removing trailing slash from `jPath` in the same place where it's removed from `tagName`.

# Type
Please mention the type of PR

* [x]Bug Fix
* [ ]New Feature

### DIFF
--- a/spec/updateTag_spec.js
+++ b/spec/updateTag_spec.js
@@ -121,6 +121,8 @@ describe("XMLParser updateTag ", function() {
                 <img width="500" height="500">
             </content>
             <script></script>
+            <lorem/>
+            <ipsum/>
         </body>
     </html>`;
         const expected = {
@@ -167,6 +169,10 @@ describe("XMLParser updateTag ", function() {
                     if(attrs.width > 200 || attrs.height > 200) return false;
                 }else if(tagname === "content"){
                     return "div"
+                }else if(tagname === "lorem") {
+                    return false;
+                }else if(jPath === "html.body.ipsum") {
+                    return false;
                 }
                 return tagname;
             },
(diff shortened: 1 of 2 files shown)


m-radzikowski:
Hey @amitguptagwl, any chance you could release a new version with this fix?

amitguptagwl:
Done. Sorry for the delay.

Issue #585, opened 2023-06-07
XMLBuilder: Attributes with `undefined` values produce odd whitespace when used with `format: true`


FiniteLooper:
* [X] Are you running the latest version?
* [X] Have you included sample input, output, error, and expected output? 
* [X] Have you checked if you are using correct configuration?
* [X] Did you try online [tool](https://naturalintelligence.github.io/fast-xml-parser/)?

### Description
Using `XMLBuilder` when the `format: true` option is on and an array contains a property with `undefined` as the value it produces odd whitespace formatting.  The text is not on the same line as the opening XML tag, and there are several spaces added before the closing tag.

The XML output is correct in how it creates the tags and their attributes, but the whitespace is just all wrong.

### Code
```js
const xmlObj = {
  "list": {
	"item":[
	  "one",
	  {"#text": "two"},
	  {"#text": "three", "@_attr": undefined},
	  {"#text": "two",   "@_attr": "foo"}
	]
  }
};
const opts = {
  ignoreAttributes: false,
  format: true
};
const builder = new fxp.XMLBuilder(opts);
console.log(builder.build(xmlObj));
```

### Actual Output
```xml
<list>
  <item>one</item>
  <item>two</item>
  <item>
three  </item>
  <item attr="foo">two</item>
</list>
```

### Expected Output
```xml
<list>
  <item>one</item>
  <item>two</item>
  <item>three</item>
  <item attr="foo">two</item>
</list>
```

What I am working on has optional object properties, so I just pass the data to the property as is.  Here is my real use case in my TypeScript project
```ts
interface IObj {
  value: string;
  part?: string;
}

function getXmlNodes(dataArr: IObj[]){
  const xmlObjs = [];
  
  for(let item of dataArr){
    xmlObjs.push({
      '#text': item.value,
      '@_part': item.part,
    });
  }

  return xmlObjs;
}
```

**Would you like to work on this issue?**

- [ ] Yes
- [X] No

cecia234:
Hi @amitguptagwl, I found the same odd behaviour. Moreover, there is another "odd" case with formatting enabled when attributes are specified as null.

Using the same structure as the one of @FiniteLooper, with the starting object as follows
```javascript
const xmlObj = {
  "list": {
	"item":[
	  "one",
	  {"#text": "two"},
	  {"#text": "three", "@_attr": undefined},
	  {"#text": "two",   "@_attr": "foo"}
	]
  }
};
```
the final output is 
```xml
[...]

PR #598, opened 2023-07-22
Fix for null and undefined attributes when building xml (#585)

cecia234:
# Purpose / Goal
After the fix `null` and `undefined` attributes in the starting object do not cause odd whitespaces and newlines in the built xml, regardless of whether `format` is `true` or `false`. Issue #585 

Perf test on XMLBuilder:

```
Running Suite: XML Builder benchmark
fxp : 76193.36948123561 requests/second
fxp - preserve order : 238552830.29948822 requests/second
xml2js  : 21083.181183955512 requests/second
```

# Type
Please mention the type of PR

* [x]Bug Fix
* [ ]New Feature

### DIFF
--- a/spec/j2x_spec.js
+++ b/spec/j2x_spec.js
@@ -483,4 +483,117 @@ describe("XMLBuilder", function() {
         expect(result).toEqual(expected);
     });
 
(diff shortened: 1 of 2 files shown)


amitguptagwl:
Thanks for your PR. Let me review it.

PR #645, opened 2024-03-07
Add support for parsing HTML numeric entities

DerZade:
# Purpose / Goal

This PR adds a `htmlNumericEntities` option which adds support for parsing HTML numeric entities.

# Type
Please mention the type of PR

* [ ] Bug Fix
* [x] New Feature

# Benchmarks

## Before
```
fxp v3 : 43885.453114704054 requests/second
fxp : 28444.884306095257 requests/second
fxp - preserve order : 35582.44370634719 requests/second
xmlbuilder2 : 12279.374624117605 requests/second
xml2js  : 17997.982194339696 requests/second
```

## After
```
fxp v3 : 56829.18072321323 requests/second
fxp : 35711.03397186936 requests/second
fxp - preserve order : 35048.53534051172 requests/second
xmlbuilder2 : 12565.714384084644 requests/second
xml2js  : 17898.357168570532 requests/second
```

### DIFF
--- a/docs/v4/5.Entities.md
+++ b/docs/v4/5.Entities.md
@@ -132,6 +132,8 @@ Following HTML entities are supported by the parser by default when `htmlEntitie
 | ₹      | Indian Rupee               | `&inr;`       | `&#8377;`        |
 ---
 
+In addition, [numeric character references](https://html.spec.whatwg.org/multipage/syntax.html#syntax-charref) are also supported. Both decimal (`num_dec`) and hexadecimal(`num_hex`).
+
 In future version of FXP, we'll be supporting more features of DOCTYPE such as `ELEMENT`, reading content for an entity from a file etc.
 
 ## External Entities
--- a/spec/entities_spec.js
+++ b/spec/entities_spec.js
@@ -377,6 +377,41 @@ describe("XMLParser Entities", function() {
         expect(result).toEqual(expected);
     });
 
+    
+    it("should parse HTML numeric entities when htmlEntities:true", function() {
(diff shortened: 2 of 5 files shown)


amitguptagwl:
Thanks for your PR

I've a few suggestions
1. I believe we don't need another option. It can be done with `htmlEntities` flag only.
2. we can add an extra  property to htmlEntities object, say `num` or `hex`. And the value can be set to a function. In the loop, we can check the type if it is string replace it as today. Otherwise, run the function as per setting in the object. It'll give the chance to disable it easily.
3. both expressions (Eg `/&#([0-9]+);/`) has `+` that make it unsafe to match very long string and will impact performance too.

Let me know your thoughts

DerZade:
@amitguptagwl I just force pushed and updated my code to reflect your suggestions:

- Removed `+` from both expressions. Decimal is now limited to 7 chars and hexadecimal to 6 chars. (The max code point is `U+10FFFF` or `1,114,112` in decimal)
- Removed the extra `htmlNumericEntities` option 
- Added both regex and replacers to `htmlEntities`-object. I choose `num_dec` and `num_hex` as keys, since there is a named char reference `num;`
- I didn't even have to check the type for string / funcion as you suggested, since [`String.prototype.replace` already supports functions as the replacer value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#replacement).

amitguptagwl:
It's live now. You can see the entry in change logs

DerZade:
Thank you so much!

PR #658, opened 2024-06-15
chore: add trend image

cwtuan:
Thank you for maintaining the `fast-xml-parser` package. It has been an invaluable tool in my projects, and I truly appreciate the hard work you and the contributors have dedicated to its development. 

As an user of this package, I would like to suggest a small enhancement to the README that I believe could benefit potential users. Specifically, I propose adding a trend image. It will visually show the increasing popularity of the `fast-xml-parser` package over recent years.

New users often look for signs of a well-maintained and widely-used package. A positive download trend can be a persuasive factor. Also, highlighting the growth of the package can encourage more contributions from the community.

Thank you for considering my suggestion.

Here's a preview:

image


amitguptagwl:
Thanks for the PR. no doubt it can be helpful but the download trends is already present on npm so it'll be helpful for only Github users. Currently, there are lot of contents on the README. And I don't want to push the contents down more. So I was wondering to reduce the size of the graph and move it to graph section.

Issue #666, opened 2024-08-04
Feature Request: option to ignore attributes by name (by array of string | regex)


mav-rik:
### Description

Please consider adding an option to the parser/builder to exclude certain attributes by name. This could be implemented as an array of attribute names or a RegExp pattern, similar to existing `include`/`exclude` API patters for paths (e.g. `typescript.json` > `exclude`).

### Use Case

We need to compare two XML documents by parsing and converting them into normalized XML outputs. Sometimes, one XML might contain extra attributes that are irrelevant for our comparison. If we exclude specific attributes during parsing, we'll easily compare the results then.

### Suggested API:
```js
const parser = new XMLParser({
  ignoreAttributes: false,
  excludeAttributes: [
    'attr-to-skip',
    /^ns:/,
    // more patterns or attributes here...
  ]
});
```

### Expected Result:
The parser should ignore the specified attributes, ensuring they do not appear in the resulting XML tree.

amitguptagwl:
this is a good suggestion. We can probably replace ignoreAttributes with this new property.

mav-rik:
> this is a good suggestion. We can probably replace ignoreAttributes with this new property.

Do you need any help with PR? I can work on `ignoreAttributes` to accept an array of regex/strings and send a PR

amitguptagwl:
Thanks @mav-rik . I'm busy with a new open source project. So it'll be helpful if you can raise a PR.

I'm just thinking from release perspective if we support both properties for sometime with deprecation warning otherwise it'll be a breaking change.

mav-rik:
> I'm just thinking from release perspective if we support both properties for sometime with deprecation warning otherwise it'll be a breaking change.

Why don't just keep backwards compatible `ignoreAttributes` prop like this:

```ts
ignoreAttributes: boolean | (string | Regex)[]
```

So when it is `true` or `false` - the current logic works.
And when it is an array of strings/regexp - then we parse attrs and filter those out based on the list.

In this case it won't be a breaking change.

mav-rik:
@amitguptagwl hey can you please take a look PR #668

Codeclimate fails with error "Avoid deeply nested control flow statements". The thing is I didn't add nor remove any of the if/else statements, I only added a few commands inside. In order to resolve Codeclimate error I'll have to refactor the whole thing which I'd like to avoid.

amitguptagwl:
[...]

PR #668, opened 2024-08-21
feat #666: add selective ignoreAttributes by pattern or callback

mav-rik:
# Purpose / Goal
Introduce a new feature that allows to selectively ignore XML attributes during parsing or building, based on an array of strings, regular expressions, or a callback function. No breaking changes, the old values `ignoreAttributes: true` and `ignoreAttributes: false` are processed same way as before.

Implements [Issue #666](https://github.com/NaturalIntelligence/fast-xml-parser/issues/666).

# Type

* [ ] Bug Fix
* [x] New Feature

### DIFF
--- a/docs/v4/2.XMLparseOptions.md
+++ b/docs/v4/2.XMLparseOptions.md
@@ -280,24 +280,95 @@ FXP by default parse XMl entities if `processEntities: true`. You can set `htmlE
 
 ## ignoreAttributes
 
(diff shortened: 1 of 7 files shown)


mav-rik:
@amitguptagwl I don't get those two codeclimate issues pointing to "Avoid deeply nested control flow statements."
It looks like those issues are not caused by my changes, those were there initially.

mnepoxuyy (approved):

PR #728, opened 2025-02-27
fix #593: support line numbers

### LINKED ISSUE #593: Feature Request: Allow including node metadata (startIndex) in output
* [X] Are you running the latest version?
* [X] Have you included sample input, output, error, and expected output? 
* [X] Have you checked if you are using correct configuration?
* [X] Did you try online [tool](https://naturalintelligence.github.io/fast-xml-parser/)?

### Description

Feature request to have a way to include metadata such as `startIndex` in the parser output

Potentially related: https://github.com/NaturalIntelligence/fast-xml-parser/issues/345

### Background

I'm working on a project that uses `fast-xml-parser` to parse XML files, it needs to report where a certain string was found in a given file. To support this I'm currently monkeypatching `XmlNode`: https://github.com/ewlsh/tree-gettext/blob/main/src/languages/xml.ts#L11

-  [X] Yes
-  [X] No

I am open to contributing but am submitting this feature request first to gather input.

srl295:
- add a new symbol, accessible via XMLParser.getStartIndexSymbol() which reflects the start index of a tag
- copy that start index in the compressed form as well

# Purpose / Goal

For our XML based parser, we want to show the line number of where an errant XML element was found.

This PR adds a symbol to objects in the output tree with the start index (in chars) of the element.

Fixes #593

# Type
Please mention the type of PR

* [ ] Bug Fix
* [X] New Feature

# Benchmark

## `dev`

```shell
$ node benchmark/XmlParser.js
fxp v3 : 46324.79199280721 requests/second
fxp : 31572.648503058208 requests/second
fxp - preserve order : 35427.55739122577 requests/second
xmlbuilder2 : 10801.649626280605 requests/second
xml2js  : 6708.125756416028 requests/second
```

## my branch

```shell
$ node benchmark/XmlParser.js
fxp v3 : 46066.87308877694 requests/second
fxp : 29175.730599200317 requests/second
fxp - preserve order : 32111.006246196102 requests/second
xmlbuilder2 : 11720.013830644504 requests/second
xml2js  : 8301.685329713537 requests/second
```

### DIFF
--- a/docs/v4/1.GettingStarted.md
+++ b/docs/v4/1.GettingStarted.md
@@ -33,4 +33,21 @@ if(XMLValidator.validate()){
 }
 ```
 
+## Start index
+
+You can get the 'start index' which is the character offset of a particular node.
+This is not available for nodes that appear as strings.
+
+```js
+const {XMLParser} = require('fast-xml-parser');
+
+const parser = new XMLParser({ignoreAttributes: false});
+const jsonObj = parser.parse(`<root><thing name="zero"/><thing name="one"/></root>`);
+const START_INDEX = parser.getStartIndexSymbol();
+// get the char offset of the start of the tag for <thing name="zero"/>
+const thingZero = jsonObj.root.thing[0];
+const thingZeroIndex = thingZero[START_INDEX]; // 6
+```
+
+
 [> Next: XmlParser](./2.XMLparseOptions.md)
--- a/spec/startIndex_spec.js
+++ b/spec/startIndex_spec.js
@@ -1,0 +1,28 @@
+
+const {XMLParser, XMLBuilder, XMLValidator} = require("../src/fxp");
+
+describe("XMLParser", function() {
+
+    it("should support the START_INDEX symbol ", function(){
+        const xmlData = `<root><foo/><bar type="quux"/><bar type="bat"/></root>`;
+        const expected = { 
+            root: {
+                foo: '',
+                bar: [
+                    { "@_type": 'quux'},
+                    { "@_type": 'bat'},
+                ],
+            }
+         };
+        
+        const parser = new XMLParser({preserveOrder:false, ignoreAttributes: false});
+        const START_INDEX = parser.getStartIndexSymbol();
+        const result = parser.parse(xmlData);
+        // console.dir(result, {depth:Infinity});
+        expect(result).toEqual(expected);
+        expect(result.root[START_INDEX]).toEqual(0); // index of <root>
+        // expect(result.root.foo[START_INDEX]).toEqual(6); // index of <foo/> - but there's no object, just a string.
+        expect(result.root.bar[0][START_INDEX]).toEqual(12);
+        expect(result.root.bar[1][START_INDEX]).toEqual(30);
+    });
+});
--- a/src/fxp.d.ts
+++ b/src/fxp.d.ts
@@ -88,6 +88,8 @@ export class XMLParser {
    * @param entityValue {string} Eg: '\r'
    */
   addEntity(entityIndentifier: string, entityValue: string): void;
+  /** Returns a Symbol that can be used to extract the node start index. */
+  getStartIndexSymbol() : Symbol;
 }
 
 export class XMLValidator{
--- a/src/xmlparser/OrderedObjParser.js
+++ b/src/xmlparser/OrderedObjParser.js
@@ -14,6 +14,7 @@ const regx =
 //const tagsRegx = new RegExp("<(\\/?)((\\w*:)?([\\w:\\-\._]+))([^>]*)>([^<]*)("+cdataRegx+"([^<]*))*([^<]+)?","g");
 
 class OrderedObjParser{
+  static START_INDEX = Symbol("Start Index of XML Node");
   constructor(options){
     this.options = options;
     this.currentNode = null;
@@ -226,7 +227,7 @@ const parseXml = function(xmlData) {
           if(tagData.tagName !== tagData.tagExp && tagData.attrExpPresent){
             childNode[":@"] = this.buildAttributesMap(tagData.tagExp, jPath);
           }
-          currentNode.addChild(childNode);
+          currentNode.addChild(childNode, OrderedObjParser.START_INDEX, i);
 
         }
 
@@ -293,6 +294,8 @@ const parseXml = function(xmlData) {
           currentNode = this.tagsNodeStack.pop();
         }
 
+        const startIndex = i;
+
         if (this.isItStopNode(this.options.stopNodes, jPath, tagName)) { //TODO: namespace
           let tagContent = "";
           //self-closing tag
@@ -313,6 +316,7 @@ const parseXml = function(xmlData) {
           }
 
           const childNode = new xmlNode(tagName);
+
           if(tagName !== tagExp && attrExpPresent){
             childNode[":@"] = this.buildAttributesMap(tagExp, jPath);
           }
@@ -323,7 +327,7 @@ const parseXml = function(xmlData) {
           jPath = jPath.substr(0, jPath.lastIndexOf("."));
           childNode.add(this.options.textNodeName, tagContent);
           
-          currentNode.addChild(childNode);
+          currentNode.addChild(childNode, OrderedObjParser.START_INDEX, startIndex);
         }else{
   //selfClosing tag
           if(tagExp.length > 0 && tagExp.lastIndexOf("/") === tagExp.length - 1){
@@ -343,7 +347,7 @@ const parseXml = function(xmlData) {
               childNode[":@"] = this.buildAttributesMap(tagExp, jPath);
             }
             jPath = jPath.substr(0, jPath.lastIndexOf("."));
-            currentNode.addChild(childNode);
+            currentNode.addChild(childNode, OrderedObjParser.START_INDEX, startIndex);
           }
     //opening tag
           else{
@@ -353,7 +357,7 @@ const parseXml = function(xmlData) {
             if(tagName !== tagExp && attrExpPresent){
               childNode[":@"] = this.buildAttributesMap(tagExp, jPath);
             }
-            currentNode.addChild(childNode);
+            currentNode.addChild(childNode, OrderedObjParser.START_INDEX, startIndex);
             currentNode = childNode;
           }
           textData = "";
--- a/src/xmlparser/XMLParser.js
+++ b/src/xmlparser/XMLParser.js
@@ -53,6 +53,11 @@ class XMLParser{
             this.externalEntities[key] = value;
         }
     }
+
+    /** get the symbol used for the start index on nodes */
+    getStartIndexSymbol() {
+        return OrderedObjParser.START_INDEX;
+    }
 }
 
 module.exports = XMLParser;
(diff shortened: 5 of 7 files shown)


amitguptagwl:
Thanks for your PR. This seems interesting. But I would have to understand it in detail.
`dev` branch is not up to date. and npm doesn't allow me to release on v4 anymore. So we would have to stick on v5 only.

srl295:
> Thanks for your PR. This seems interesting. But I would have to understand it in detail.
> 
> `dev` branch is not up to date. and npm doesn't allow me to release on v4 anymore. So we would have to stick on v5 only.

Ok. Does v5 have this same api?

Master branch would not install, "release-please" failed.

srl295:
You can see how I use it in the linked commit

amitguptagwl:
A few thoughts on this feature.
- What problem would it solve? Or rather what are the use cases of this feature?
- Is it optional to the user?
- What would be the impact on memory?
- How would it impact old node versions and browsers?
-Please check [v6](https://github.com/NaturalIntelligence/fast-xml-parser/tree/master/src/v6) as well.

srl295:
> A few thoughts on this feature.
> 
> - What problem would it solve? Or rather what are the use cases of this feature?

For a compiler it's very helpful to users to know what line an error was. So we can say "your element some thing had an error on line 44" 


> 
> - Is it optional to the user?

Might be a good idea if it impacts perf or memory 

> - What would be the impact on memory?

Adds a symbol on each node 

> 
> - How would it impact old node versions and browsers?

Should be ok if they have Symbol

> 
> -Please check [v6](https://github.com/NaturalIntelligence/fast-xml-parser/tree/master/src/v6) as well.

Ok. Is it ready to use ?

amitguptagwl:
v6 is for experimental use. It is designed to overcome the design restrictions of current version. It is being difficult to expand current version for new features and customization. I want user have more controller on parser and builder.

If our use case is just to tell the exact line number in case of error, there is no point to store that data after parsing. I believe I've tied to cover in v6 if I remember correctly.

In my knowledge, Symbols are not backward compatible. So I afraid we can't go ahead with them at least in current major version.

srl295:
But I am not talking about an XML error but an application error. 

My XML document specifies a keyboard. I want to tell the user, your keyboard is malformed because it has too many keys on the row. So I want to say, the xml file line 223 has an error. 

But the document parses fine.

amitguptagwl:
In your case, you want to preserve that error information even after parsing?

srl295:
@amitguptagwl let me be clear what 'error' means here.

1. I have a file specifying keyboards, for example:

```xml
  <keyboard3>
    <!--  .. other parts of the file ... -->
     <layers formId="us">
        <layer modifiers="shift">
            <row keys="C-tikka bang at euro dollar percent caret amp asterisk open-paren close-paren underscore plus" />
            <row keys="Q W E R T Y U I O P G-tikka H-maqtugha Z-tikka" />
            <row keys="A S D F G H J K L colon double-quote" />
            <row keys="Z X C V B N M open-angle close-angle question" />
            <row keys="space" />
            <row keys="xxx yyy zzz" /> <!-- This row shouldn't be here, too many rows <<<<<<<< -->
        </layer>
      </layers>
     <!-- .. other parts of the file ... -->
  </keyboard3>   
```

2. The application parses the file

```js
   const xml = readFileSync('file.xml', 'utf-8');
   const { XMLParser} = require('fast-xml-parser');
   const obj = new XMLParser({ignoreAttributes: false}).parse(xml);
```

3. Next, the application processes `obj` and discovers that the `<layer>` object has too many lines  for the `us` layout. (has 6, max 5).  This is not a parsing error - the XML is well-formed. But, it's an error within the application.

4. The application wants to give the user a more helpful error message- the standard type of message that any compiler would give.

```shell
file.xml:14:2: Error: layer has 6 rows, max 5 allowed.
```

5. The user can navigate to line 14 and correct the application-level error.

amitguptagwl:
If I understand you correctly, are you looking for validator based on business requirement? Have a look on [detailed-xml-validator](https://www.npmjs.com/package/detailed-xml-validator)?

srl295:
> If I understand you correctly, are you looking for validator based on business requirement? Have a look on [detailed-xml-validator](https://www.npmjs.com/package/detailed-xml-validator)?

No, it's not a validator… it's a compiler.

I want to preserve the line numbers in the object tree, so that much, much later, I can tell the user where the original XML object came from.

srl295:
Let me give another example.

Take https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/spec/assets/sample.xml#L14

Imagine an application that wants to print out:

```shell
$ checkBirthdays sample.xml
sample.xml:14: Today is your <birthday>!
```

So that someone can see that line 14 of sample.xml had the `<birthday>` element.

If `parse()` returns a tree that includes the startIndex of 454 on the `birthday`, then the application can calculate that it is line 14.

amitguptagwl:
Apologies for this deep discussion and more questions for better understanding.
- In case of multiple same tags, what your application should return? line number of first tag?
- If I understand correctly, you're associating line number with each JSON property which was a tag. So how would you pass that property to query?
- As per the  shared examples, you would preserve XML and parsed output (i.e. JS object) both. What if I search for that tax just as normal text, how would it be different?

srl295:
> Apologies for this deep discussion and more questions for better understanding.

No problem! Thanks for being willing to discuss. Also note my other PR #729 which is based on the master (5.x) branch.  My examples here are from that PR (but i'm answering here because that's where you are replying)

Here's another code snippet:

```
> q = new XMLParser({preserveStartIndex:true, ignoreAttributes:false}).parse(`<xml><item name="one"/><item name="two"/><item name="three"/></xml>`);
> console.dir(q, {depth: Infinity})
{
  xml: {
    item: [
      { '@_name': 'one', [Symbol(Start Index of XML Node)]: 5 },
      { '@_name': 'two', [Symbol(Start Index of XML Node)]: 23 },
      { '@_name': 'three', [Symbol(Start Index of XML Node)]: 41 }
    ],
    [Symbol(Start Index of XML Node)]: 0
  }
}
```

You see that the 'start index' is present on the parsed tree, but as a symbol (so it won't show up in `Object.keys`.

> * In case of multiple same tags, what your application should return? line number of first tag?

Well, that's up to the application. I'm not searching for the XML tag, it's present in the result of parse().  It might return a message about every line. (See the code snippet below)

> * If I understand correctly, you're associating line number with each JSON property which was a tag. So how would you pass that property to query?

I don't see what query has to do with this use case. I'm not running a query on the object output. But probably a query would ignore the symbol attribute.

> * As per the  shared examples, you would preserve XML and parsed output (i.e. JS object) both. What if I search for that tax just as normal text, how would it be different?

I think you are talking about this case:

```shell
> q = new XMLParser({preserveStartIndex:true, ignoreAttributes:false, preserveOrdering: true}).parse(`<xml><item name="one"/><item name="two"/><item name="three"/><foo>bar</foo></xml>`);
{
  xml: {
    item: [ [Object], [Object], [Object] ],
    foo: 'bar',
    [Symbol(Start Index of XML Node)]: 0
  }
}
```

in this case, the object `<foo>bar</foo>` doesn't get a symbol in it because it's just a string.  Can't have properties on strings.


However, you can use `preserveOrder:true`:

```shell
> q = new XMLParser({preserveStartIndex:true, ignoreAttributes:false, preserveOrder: true}).parse(`<xml><item name="one"/><item name="two"/><item name="three"/><foo>bar</foo></xml>`);
[
  {
    xml: [ [Object], [Object], [Object], [Object] ],
    [Symbol(Start Index of XML Node)]: 0
  }
]
> console.dir(q, {depth: Infinity})
[
  {
    xml: [
      {
        item: [],
        ':@': { '@_name': 'one' },
        [Symbol(Start Index of XML Node)]: 5
      },
      {
        item: [],
        ':@': { '@_name': 'two' },
        [Symbol(Start Index of XML Node)]: 23
      },
      {
        item: [],
        ':@': { '@_name': 'three' },
        [Symbol(Start Index of XML Node)]: 41
      },
      {
        foo: [ { '#text': 'bar' } ],
        [Symbol(Start Index of XML Node)]: 61
      }
    ],
    [Symbol(Start Index of XML Node)]: 0
  }
]
```

amitguptagwl:
It is not necessary that the XML is well formatted, it can be minimized as well. Let's understand the need with examples. Eg
```xml
<Person>
   <birthday>
     20 May 1993
   </birthday>
</Person>
```

For an application which is interested to find a line number with tag's start position, may result undesirable result.
```
$ checkBirthdays sample.xml
sample.xml:2: <birthday>
```

Sometimes, a tag can be broken in multiple lines.
---

Let's take another example of validation. Suppose, I want to validate on the frequency of tags and report the extra tags
```xml
  <keyboard3>
    <!--  .. other parts of the file ... -->
     <layers formId="us">
        <layer .../> <!-- other layers -->
        <layer modifiers="shift">
            <row keys="C-tikka bang at euro dollar percent caret amp asterisk open-paren close-paren underscore plus" />
            <row keys="Q W E R T Y U I O P G-tikka H-maqtugha Z-tikka" />
            <row keys="A S D F G H J K L colon double-quote" />
            <row keys="Z X C V B N M open-angle close-angle question" />
            <row keys="space" />
            <row keys="xxx yyy zzz" /> <!-- This row shouldn't be here, too many rows <<<<<<<< -->
            <row keys="xxx yyy zzz" /> <!-- This row shouldn't be here, too many rows <<<<<<<< -->
            <row keys="xxx yyy zzz" /> <!-- This row shouldn't be here, too many rows <<<<<<<< -->
            <row keys="xxx yyy zzz" /> <!-- This row shouldn't be here, too many rows <<<<<<<< -->
            <row keys="xxx yyy zzz" /> <!-- This row shouldn't be here, too many rows <<<<<<<< -->
        </layer>
      </layers>
     <!-- .. other parts of the file ... -->
  </keyboard3>   

```

If I define a rule file
```xml
<keyboard3>
    <!--  .. other parts of the file ... -->
     <layers>
        <layer repeatable minOccurs="1">
            <row repeatable minOccurs="5" maxOccurs="5">
            <:a>
                 <keys nillable=false>
            </:a>
            </row>
        </layer>
      </layers>
     <!-- .. other parts of the file ... -->
  </keyboard3>   
```

Response
```
[{ code: 'maxLength', path: 'keyboard3.layers.layer[0]', actual: '10', expected: 5 }]
```

Validator already has provision for custom response. And cn be upgraded to collect extra information.

srl295:
@amitguptagwl I'm not asking about a validator.  this is after the parsing phase.

Did you see my example above?

```js
parsed = {
  xml: {
    item: [
      { '@_name': 'one', [Symbol(Start Index of XML Node)]: 5 },
      { '@_name': 'two', [Symbol(Start Index of XML Node)]: 23 },
      { '@_name': 'three', [Symbol(Start Index of XML Node)]: 41 }
    ],
    [Symbol(Start Index of XML Node)]: 0
  }
}
```

I can get the offsets of the three items as follows:

1. parsed.xml.item[0][XMLParser.getStartIndexSymbol()]
2. parsed.xml.item[1][XMLParser.getStartIndexSymbol()]
3. parsed.xml.item[2][XMLParser.getStartIndexSymbol()]

Just as I can do parsed.xml.item[0]['@_name'] to get the name

amitguptagwl:
We discussed 2 examples and I didn't find the need of this feature for those use cases at least. This might be frustrating for you but I would need more examples to understand the need. 

As you commented about after parsing approach , I also need to understand how implementation would impact a feature?

PR #797, opened 2026-03-07
fix: stopNodes should preserve text content as string (#795)

### LINKED ISSUE #795: stopNodes still parses numbers
The output of the following is `"number"`, when according to the documentation it should return a string.

```js
import { XMLParser } from "fast-xml-parser";

	stopNodes: ["*.a"]
});

const text = `<a>6</a>`;

console.log(typeof parser.parse(text).a);
```

abhu85:
## Summary
- Fixes stopNodes content being parsed as numbers/booleans instead of remaining as strings
- `<a>6</a>` with `stopNodes: ["*.a"]` now returns `{ a: "6" }` instead of `{ a: 6 }`

## Problem
When using `stopNodes` option, numeric and boolean-like text content was being converted to JavaScript numbers/booleans. According to the documentation, `stopNodes` should preserve content as raw text.

**Reproduction (from #795):**
```js
const parser = new XMLParser({ stopNodes: ["*.a"] });
console.log(typeof parser.parse("<a>6</a>").a); // "number" - should be "string"
```

## Solution
Added `doNotParseTagValue` parameter to `parseTextData` function and set it to `true` when processing stopNode content. This skips the `parseValue` call that converts strings to numbers/booleans, while still allowing `tagValueProcessor` to work for user customization.

## Test Plan
- [x] All 282 existing tests pass
- [x] Added regression test covering: numeric strings, boolean strings ("true"/"false"), negative/decimal numbers, scientific notation

## Compatibility
- [x] No breaking changes
- [x] Backwards compatible - existing `tagValueProcessor` behavior preserved

Fixes #795

### DIFF
--- a/spec/stopNodes_spec.js
+++ b/spec/stopNodes_spec.js
@@ -448,11 +448,38 @@ describe("XMLParser StopNodes", function () {
     }
     const parser = new XMLParser(options);
     // console.log(JSON.stringify(parser.parse(xml)));
-    
+
     let result = parser.parse(xmlData);
 
     // console.log(JSON.stringify(result,null,4));
     expect(result).toEqual(expected);
 
+  });
+  it("should preserve text content as string, not parse as number/boolean (#795)", function() {
+    const parser = new XMLParser({ stopNodes: ["*.a"] });
+
+    // Numeric string should remain string
+    let result = parser.parse("<a>6</a>");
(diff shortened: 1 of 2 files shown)


amitguptagwl:
Hi @abhu85 ,
I'm working on a major change. And I have fixed a lot of part of this in that. I would update you soon and hoping to release that tonight or tomorrow morning.

PR #818, opened 2026-04-14
fix: stop node namespace prefix parameter ordering in readStopNodeData

iruizsalinas:
# Purpose / Goal

Fix incorrect parameter ordering in `readStopNodeData` that breaks stop node tracking when the tag has a namespace prefix.

# Type
* [x] Bug Fix

`readStopNodeData` calls `readTagExp(xmlData, i, '>')` at line 760, but `'>'` ends up in the `removeNSPrefix` parameter instead of `closingChar`. Since it's truthy, namespace prefixes get stripped from opening tags inside the stop node, while closing tags keep them. This throws off `openTagCount` because opens and closes no longer match.

```js
const parser = new XMLParser({ stopNodes: ["root.ns:code"] });
const result = parser.parse(
  "<root><ns:code>safe <ns:code>nested</ns:code> still raw</ns:code></root>"
);
// result.root["ns:code"] should be "safe <ns:code>nested</ns:code> still raw"
// but instead the stop node exits early at the inner </ns:code>
```

The fix is changing `readTagExp(xmlData, i, '>')` to `readTagExp(xmlData, i, false)` since `closingChar` already defaults to `">"`.

### DIFF
--- a/src/xmlparser/OrderedObjParser.js
+++ b/src/xmlparser/OrderedObjParser.js
@@ -757,7 +757,7 @@ function readStopNodeData(xmlData, tagName, i) {
         const closeIndex = findClosingIndex(xmlData, "]]>", i, "StopNode is not closed.") - 2;
         i = closeIndex;
       } else {
-        const tagData = readTagExp(xmlData, i, '>')
+        const tagData = readTagExp(xmlData, i, false)
 
         if (tagData) {
           const openTagName = tagData && tagData.tagName;


amitguptagwl:
Thanks for your effort and time. Let me test it with some tests and then I'll check the PR. Probably on coming monday

amitguptagwl:
As you have not mentioned the version, I'm assuming it is latest. And I believe, you're using wrong syntax for namespace then. Application supports [path-expression-matcher](https://github.com/NaturalIntelligence/path-expression-matcher).

Can you please check with correct syntax and let me know if you're still facing the issue.

iruizsalinas:
Hello, I'm on v5.7.1. Just tested with `::` too (`stopNodes: ["root.ns::a"]` + `removeNSPrefix: true`) and it breaks the same way whenever the stopnode has a nested same-name tag:

`<root><ns:a><ns:a>x</ns:a></ns:a></root>` gives me `"<ns:a>x"` instead of the full content, with either syntax.

The bug isn't in the matching, it's in `readStopNodeData`. `'>'` was being passed as the `removeNSPrefix` arg to `readTagExp` instead of `closingChar`, so nested opens get stripped while closes stay raw.

amitguptagwl:
Somehow a related test was passing so I raised the doubt. As I'm going on Holiday for few days. I'll merge, and release after that.

iruizsalinas:
Thanks, have a good one.

amitguptagwl:
change are published. please check once. thanks for your time and effort

PR #826, opened 2026-04-29
fix: resolve numeric character reference external entities

### LINKED ISSUE #825: v5.7.x rejects valid XML numeric character references like &#xD;
## Description

After upgrading from `fast-xml-parser@5.5.8` to `5.7.1`, the parser rejects valid XML numeric character references such as `&#xD;` (carriage return).

## Error Message

```
Error: [EntityReplacer] Invalid character '#' in entity name: "#xD"
```

## Reproduction

This occurs when parsing XML responses from AWS S3 (via `@aws-sdk/xml-builder` which uses `fast-xml-parser`). AWS S3 responses contain XML with standard numeric character references like `&#xD;`.

## Expected Behavior

`&#xD;` is a valid XML numeric character reference representing carriage return (U+000D). According to the XML specification, numeric character references in the form `&#xNNNN;` (hexadecimal) or `&#NNNN;` (decimal) should be accepted.

Reference: https://www.w3.org/TR/xml/#NT-CharRef

## Root Cause Hypothesis

The issue appears to be related to the dependency change in v5.7.x from `fast-xml-builder` to `@nodable/entities`. The new entity handling seems to incorrectly treat `#xD` as an entity name rather than recognizing it as a numeric character reference.

## Environment

- fast-xml-parser version: 5.7.1
- Node.js version: 24.x
- Downstream dependency: @aws-sdk/xml-builder (AWS SDK v3)

## Workaround

Pinning to `fast-xml-parser@5.5.8` resolves the issue.

abhu85:
## Summary
Fix regression where `addEntity('#xD', '\r')` and similar numeric character reference external entities are silently ignored since v5.7.x.

## Problem
The `addEntity` API documents that numeric character references can be registered as external entities (e.g., `parser.addEntity('#xD', '\r')` for `&#xD;`). This worked before v5.7.0 but broke with the migration to `@nodable/entities`:

- **v5.7.1**: `setExternalEntities()` calls `validateEntityName()` which throws `Invalid character '#' in entity name: "#xD"` 
- **v5.7.2**: The `validateEntityName` error was bypassed by passing external entities through the constructor, but `#`-prefixed entities are still silently ignored because the `EntityDecoder`'s NCR path intercepts `#`-prefixed tokens before checking the named entity map

This breaks downstream users (reported by AWS SDK users in #825).

## Solution
Separate `#`-prefixed external entities from regular named entities. Regular entities are passed to the `EntityDecoder`'s named entity map as before. Numeric (`#`-prefixed) entities are resolved via pre-processing — replacing `&#xD;`, `&#13;`, etc. with their registered values before the `EntityDecoder`'s decode pass runs.

This preserves backward compatibility:
- `addEntity('#xD', '\r')` works correctly again
- Default behavior (no `addEntity`, no `htmlEntities`) is unchanged
- `htmlEntities: true` continues to decode all numeric character references
- Custom `entityDecoder` option is unaffected

## Test Plan
- [x] Re-enabled the disabled test for `addEntity("#xD", "\r\n")`
- [x] Added tests for multiple hex (`&#xD;`, `&#x20;`, `&#x41;`) and decimal (`&#13;`) numeric character reference external entities
- [x] All 316 existing tests pass
- [x] Verified with AWS SDK XML response patterns

Fixes #825

### DIFF
--- a/spec/entities_spec.js
+++ b/spec/entities_spec.js
@@ -537,13 +537,39 @@ describe("XMLParser External Entities", function () {
         }).toThrowError("An entity must be set without '&' and ';'. Eg. use '#xD' for '&#xD;'");
     });
 
-    xit("should set and parse for valid entity set externally", function () {
+    it("should set and parse for valid entity set externally", function () {
         const xmlData = `<note>&unknown;&#xD;last</note> `;
 
         const parser = new XMLParser();
         parser.addEntity("#xD", "\r\n");
-        expect(() => parser.parse(xmlData))
-            .toThrowError(`[EntityReplacer] Invalid character '#' in entity name: "#xD"`);
+        let result = parser.parse(xmlData);
+
+        expect(result.note).toEqual(`&unknown;\r\nlast`);
+    });
+
+    it("should support multiple numeric character reference external entities", function () {
+        const xmlData = `<root>&#xD;&#13;&#x20;&#x41;</root>`;
+
+        const parser = new XMLParser();
+        parser.addEntity("#xD", "\r");
+        parser.addEntity("#13", "\r");
+        parser.addEntity("#x20", " ");
+        parser.addEntity("#x41", "A");
+        let result = parser.parse(xmlData);
+
+        expect(result.root).toEqual("\r\r A");
+    });
+
+    it("should handle hex and decimal numeric external entities together", function () {
+        const xmlData = `<root>Hello&#x20;World&#xD;&#13;End</root>`;
+
+        const parser = new XMLParser();
+        parser.addEntity("#x20", " ");
+        parser.addEntity("#xD", "\r");
+        parser.addEntity("#13", "\r");
+        let result = parser.parse(xmlData);
+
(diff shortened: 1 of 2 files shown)


amitguptagwl:
numerical entities are already internally supported. I couldn't understand the purpose of this workaround?

abhu85:
You're right — verified that numerical entities resolve correctly with `htmlEntities: true` on v5.7.2:

```js
const parser = new XMLParser({ processEntities: true, htmlEntities: true });
parser.parse('<root>test&#xD;value</root>');
// → { root: "test\rvalue" } ✓
```

`&#65;`, `&#x41;`, `&#13;` all resolve correctly as well. The pre-processing approach in this PR is unnecessary since the EntityDecoder already handles NCR when `htmlEntities` is enabled.

The original issue reporter was using `addEntity('#xD', '\r')` instead of enabling `htmlEntities: true`. Closing this PR — a comment on #825 pointing to `htmlEntities: true` would resolve the issue.

PR #832, opened 2026-05-14
Update 2.XMLparseOptions.md to fix `tagValueProcessor` example output

loucadufault:
# Purpose / Goal

Fix the docs to have the correct output. There are no such `a` attributes in the input so they are removed. Some text content therefore gets inlined. The `root`'s `checked` attribute is also added.

# Type
* [x] Bug Fix
* [ ] New Feature

### DIFF
--- a/docs/v4, v5/2.XMLparseOptions.md
+++ b/docs/v4, v5/2.XMLparseOptions.md
@@ -1298,24 +1298,15 @@ Output
 {
     "root": {
         "a": [
+            "wow",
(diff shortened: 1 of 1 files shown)


hiSandog:
Since this is a docs-output fix, it would be great to include the exact input XML and before/after parsed result in the PR description or test fixture. That makes it obvious why the removed `a` attributes and inlined text are the correct example output rather than just a formatting change.

PR #835, opened 2026-06-01
fix: preserve carriage returns inside CDATA

### LINKED ISSUE #512: Carriage Return characters (\r) are incorrectly replaced with Newline characters (\n)

### Description

Carriage Return characters (`\r`) found within parsed XML are incorrectly converted to newline characters (`\n`). The online tool does not exhibit this issue and I am uncertain as to why this is (possibly configuration options?).

I did a quick scan of the source code and found a [likely culprit](https://github.com/NaturalIntelligence/fast-xml-parser/blob/f962cb5785484986f27b643eb3ed9ad73d387c61/src/xmlparser/OrderedObjParser.js#L175):

```js
const parseXml = function(xmlData) {
  xmlData = xmlData.replace(/\r\n?/g, "\n"); //TODO: remove this line
```

That pretty clearly replaces any `\r` character (and possibly `\r\n` pair) with `\n`.

### Input

The XML that exhibits this issue is of the form:

```xml
<properties object="" engine="">
    <property type="string" name="x" state="changed">
        <![CDATA[This is a carriage return \r...]]>
    </property>
    <property type="string" name="y" state="changed">
        <![CDATA[\r]]>
    </property>
</properties>
```

### Code
The code is pretty straightforward.

```ts
const XML_OPTIONS_NO_TAG_PARSE: fastXMLParser.X2jOptionsOptional = {
    attributeNamePrefix: "@",
    ignoreAttributes: false,
    parseAttributeValue: false,
    parseTagValue: false,
    textNodeName: "#value",
};
const XML_PARSER_NO_TAG_PARSE = new fastXMLParser.XMLParser(XML_OPTIONS_NO_TAG_PARSE);

// ...

const parsed = XML_PARSER_NO_TAG_PARSE.parse(xmlData);
```

After that code runs, the parsed text node content has `\n` instead of the expected `\r`.

### Output

Running the above results in the following JSON:

```json
{
    "properties": {
        "property": [
            {
                "#value": "This is a carriage return \n...",
                "@type": "string",
                "@name": "x",
                "@state": "changed"
            },
            {
                "#value": "\n",
                "@type": "string",
                "@name": "y",
                "@state": "changed"
            },
        ],
        "@object": "",
        "@engine": ""
    }
}
```

### expected data

I expect the following output:

```json
{
    "properties": {
        "property": [
            {
                "#value": "This is a carriage return \r...",
                "@type": "string",
                "@name": "x",
                "@state": "changed"
            },
            {
                "#value": "\r",
                "@type": "string",
                "@name": "y",
                "@state": "changed"
            },
        ],
        "@object": "",
        "@engine": ""
    }
}
```

- [ ] Yes
- [x] No

viraatdas:
Fixes #512

## Problem
`parseXml` normalizes line endings with `xmlData.replace(/\r\n?/g, "\n")` at the top of `OrderedObjParser`, which also rewrites `\r` inside CDATA sections. CDATA content must be preserved byte for byte, so a literal carriage return inside CDATA comes back as `\n`.

## Fix
Replace the blanket normalization with a CDATA-aware pass that skips CDATA sections (and tags/comments) and normalizes line endings only outside CDATA. Carriage returns inside CDATA are now preserved; everything else still normalizes as before. Edits `src/xmlparser/OrderedObjParser.js` (source), not the generated bundle.

## Tests
Added a regression test ("should preserve carriage returns in CDATA") that fails before the change and passes after, plus a guard test for normalization outside CDATA. Full jasmine suite passes.

### DIFF
--- a/spec/cdata_spec.js
+++ b/spec/cdata_spec.js
@@ -383,4 +383,75 @@ patronymic</person></root>`;
         // console.log(JSON.stringify(result,null,4));
         expect(result).toEqual(expected);
     });
+
+    it("should preserve carriage returns in CDATA", function () {
+        const xmlData = `<properties object="" engine="">
+    <property type="string" name="x" state="changed">
+        <![CDATA[This is a carriage return \r...]]>
+    </property>
+    <property type="string" name="y" state="changed">
+        <![CDATA[\r]]>
+    </property>
+</properties>`;
+
+        const expected = {
+            "properties": {
+                "property": [
+                    {
+                        "#value": "This is a carriage return \r...",
+                        "@type": "string",
+                        "@name": "x",
+                        "@state": "changed"
+                    },
+                    {
+                        "#value": "\r",
+                        "@type": "string",
+                        "@name": "y",
+                        "@state": "changed"
+                    }
+                ],
+                "@object": "",
+                "@engine": ""
+            }
+        };
+
+        const options = {
+            attributeNamePrefix: "@",
+            ignoreAttributes: false,
+            parseAttributeValue: false,
+            parseTagValue: false,
+            textNodeName: "#value",
+        };
+
+        const parser = new XMLParser(options);
+        let result = parser.parse(xmlData);
+
+        expect(result).toEqual(expected);
+    });
+
+    it("should normalize carriage returns outside CDATA", function () {
+        const xmlData = `<root attr="x\ry"><text>x\ry</text><text>x\r\ny</text></root>`;
+
+        const expected = {
+            "root": {
+                "text": [
+                    "x\ny",
+                    "x\ny"
+                ],
+                "@_attr": "x\ny"
+            }
+        };
+
+        const options = {
+            ignoreAttributes: false,
+            parseAttributeValue: false,
+            parseTagValue: false,
+            trimValues: false,
+        };
+
+        const parser = new XMLParser(options);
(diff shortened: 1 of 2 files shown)


amitguptagwl:
Thanks for the PR. new line and carriage return were being removed in starting to simplify regex if I remember correctly. But now most of the code is not using regex. So I'm not sure if we need to remove them at all. Will check once

amitguptagwl:
I went through this and I felt that we need 2 more conditions to handle. Current edit suggest to handle CDATA only but the new line problem is real for other tags too like `pre` or `script` for HTML. Same with XML tags with `xml:space="preserve"`. So I was wondering if we can provide an option for the user 

```
preserveSpaceFor : ["pre","script","style"]
```

It should by default support CDATA and `xml:space="preserve"`. What is your view on this?

But this will not be backward comparible. so we'll have to introduce a flag or rather support Expressions

```
preserveSpaceFor : ["..pre","[xml:space="preserve"]"]
```

But expressions support at the time of traversing not before parsing starts. I think this need to be designed properly. However, [@nodable/flexible-xml-parser](https://github.com/nodable/flexible-xml-parser) would be easier to have this feature.

amitguptagwl:
Changes are supported in [@nodable/flexible-xml-parser](https://github.com/nodable/flexible-xml-parser)

PR #871, opened 2026-09-05
fix: lowercase doctype not recognized as DOCTYPE (fixes #870)

### LINKED ISSUE #870: ignoreDeclaration does not ignore lower case doctype declaration
* [x] Have you checked the [docs](https://github.com/NaturalIntelligence/fast-xml-parser/tree/master/docs) for helpful APIs and examples?

### Description
`ignoreDeclaration` does not ignore lower case doctype declaration.
### Code
```js
const html = `<!doctype html><html></html>`;

  ignoreDeclaration: true,
});
let jObj = parser.parse(html);

console.log(jObj);
```

### Output
```js
{ '!doctype': { html: '' } }
```

### expected data
```js
{ html: '' }
```

- [ ] Yes
- [x] No

afonsojanu:
Closes #870.

The check for a DOCTYPE tag only matched an uppercase `D` (both in the actual parser and, separately, in the input validator), so something like this:

```js
const html = `<!doctype html><html></html>`;
const parser = new XMLParser({ ignoreDeclaration: true });
parser.parse(html);
```

produced `{ '!doctype': { html: '' } }` instead of just `{ html: '' }`, because the lowercase version fell through to the normal opening-tag path and got treated as a real element.

Strict XML does require DOCTYPE to be uppercase, but a lot of what gets thrown at this library in practice is HTML rather than strict XML (this is basically what the issue's own repro does), and HTML's doctype keyword has always been case-insensitive. So I made both checks case-insensitive rather than only matching `D`.

Also fixed the same case-sensitivity gap in `src/validator.js`'s comment/CDATA/DOCTYPE tag reader, since with strict validation on (`parser.parse(data, true)`) it threw `char 'd' is not expected` on the same input before this could even reach the actual parsing step.

Added a test for the lowercase form next to the existing DOCTYPE tests in `spec/entities_spec.js`. Ran the full jasmine suite locally, 326 specs pass (2 pre-existing pending/xit specs unrelated to this).


amitguptagwl:
Thanks for your effort and time. But this PR is break backward compatibility. Hence, needs to be closed. Additionally, HTML is not the primary goal of this library thought it supports many features. So we'll have to tackle such cases through config or in https://github.com/nodable/flexible-xml-parser

